Issue #22, opened 2022-07-13
EPL-347 Error when post/unpost transactions from financial account


valeriag-etendo:
# Error's description
When trying to post/unpost a transaction from the financial account it throws the error "Table not found".
The error comes from the validation performed in the file: modules_core/com.smf.jobs.defaults/src/com/smf/jobs/defaults/Post.java (attached image), where the table is not included and the validation fails.

# Steps to reproduce the error
1. Go to Financial Account window
2. Select a Financial Account and a Transaction
4. Click on Post/Unpost.

![image-20220712-201259](https://user-images.githubusercontent.com/95085326/178747148-49495e20-96cb-4d36-9d3e-95d1abe6476a.png)

# Expected behavior
You must be able to post or reverse the transaction.

# Affected Version
Etendo 22.1.3
Etendo 22.2.0

PR #31, opened 2022-07-18
Issue #22: fix accounting for some missing tables

juabaez:
EPL-347 add a new hql query to search for missing tables in the same way as it is done in AcctServer

### DIFF
--- a/modules_core/com.smf.jobs.defaults/src/com/smf/jobs/defaults/Post.java
+++ b/modules_core/com.smf.jobs.defaults/src/com/smf/jobs/defaults/Post.java
@@ -4,14 +4,17 @@ import com.smf.jobs.Action;
 import com.smf.jobs.ActionResult;
 import com.smf.jobs.Result;
 
+import org.apache.commons.lang.StringUtils;
 import org.apache.commons.lang.mutable.MutableBoolean;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 import org.codehaus.jettison.json.JSONObject;
+import org.hibernate.query.Query;
 import org.openbravo.base.exception.OBException;
 import org.openbravo.base.secureApp.VariablesSecureApp;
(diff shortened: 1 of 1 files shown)


isaiasb-etendo (approved):

modules_core/com.smf.jobs.defaults/src/com/smf/jobs/defaults/Post.java
```diff
@@ -128,9 +131,32 @@ private String getDateProperty(String adTableId, VariablesSecureApp vars) {
...
+      /*
+       * Position 0 = ACCTDATECOLUMN
+       * Position 1 = ACCTCLASSNAME
+       * */
+      Object[] acctinfo = getTableInfo(adTableId);
+      if (acctinfo == null) {
+        throw new OBException(
+            OBMessageUtils.messageBD(new DalConnectionProvider(), "TableNotFound", vars.getLanguage()));
+      }
+      if (acctinfo.length != 0 && (!acctinfo[1].equals("") && !acctinfo[0].equals(""))) {
```
sebastianbarrozo, line 143: Will be more accurate if you check if array length is equal to 2 if you later in condition you compare index 0 and 1 with an empty string.

modules_core/com.smf.jobs.defaults/src/com/smf/jobs/defaults/Post.java
```diff
@@ -128,9 +131,32 @@ private String getDateProperty(String adTableId, VariablesSecureApp vars) {
...
+        property = StringUtils.uncapitalize((String) acctinfo[0]);
```
sebastianbarrozo, line 144: Why you get `acctclassname` and never use it? Is not relevant for this process?

sebastianbarrozo (changes requested):

sebastianbarrozo (approved):

PR #143, opened 2023-02-06
Feature ETO-91: Improve Jenkins pipelines

valeriag-etendo:


juabaez (approved):

pipelines/unittests/Jenkinsfile
```diff
@@ -106,15 +124,36 @@ pipeline {
 //___________________________________________________________________________________________
 // -------------------------------- CLEAN DATABASE ------------------------------------------
         stage('Clean Database') {
+            when {
+                expression {
+                    env.STATUSBUILD == "1"
```
RubenEtendo, line 129: Couldn't there be a situation where the build fails, but the database was created?

For example, while executing the install task, the update.database fails, but the database was already created before that

RubenEtendo (approved):

PR #171, opened 2023-04-24
Feature EPL-598: modified SequencesGenerator to allow extend

juabaez:


src/com/etendoerp/sequences/SequencesGenerator.java
etendobot, line 1: # GPT Review for SequencesGenerator.java
From the Git diff, it seems that the only change made to the code was to move the `setAdminMode()` method call above the `try` block. This change doesn't seem to have any significant impact on the code's performance or functionality.

However, if we assume that the code is part of a larger system, there may be other areas that could be optimized. Without more information, it's difficult to provide specific advice.

As for a brief code example, one potential improvement would be to use a `try-with-resources` block to automatically close resources such as `Client` and `OBContext`. This can help prevent resource leaks and improve the code's readability. Here's an example:

```
protected ActionResult action(JSONObject parameters, MutableBoolean isStopped) {
  try (OBContext ctx = OBContext.getOBContext()) {
    setAdminMode();
    //Get Current Client
    Client client = ctx.getCurrentClient();
    // Other code here
  } catch (Exception e) {
    // Handle exceptions here
  }
  return new ActionResult();
}
```

src/com/etendoerp/sequences/SequencesGenerator.java
```diff
@@ -25,118 +26,128 @@
...
+  }
+
+  public boolean existsSequence(Column column, Client client, Organization organization, DocumentType documentType) {
+    OBCriteria<Sequence> sequenceOBCriteria = OBDal.getInstance().createCriteria(Sequence.class);
+    sequenceOBCriteria.add(Restrictions.eq(Sequence.PROPERTY_CLIENT, client));
+    sequenceOBCriteria.add(Restrictions.eq(Sequence.PROPERTY_COLUMN, column));
+    if (documentType != null) {
+      sequenceOBCriteria.add(Restrictions.eq(Sequence.PROPERTY_DOCUMENTTYPE, documentType));
     }
+    sequenceOBCriteria.add(Restrictions.eq(Sequence.PROPERTY_ORGANIZATION, organization));
```
Matias-Bernal, line 145: Es mas rapido usar un setMaxResult(1), un uniqueResult y preguntar si el resultado es distinto de null asi no usas toda la tabla

Matias-Bernal (changes requested):

src/com/etendoerp/sequences/SequencesGenerator.java
isaiasb-etendo: The GitHub comments, should be in English

PR #203, opened 2023-07-18
Hotfix EPL-592: Restriction so that the current balance of bp is not …

brunoromerosmf:
…changed when a reconciliation is created and set in the transaction

### DIFF
--- a/src/org/openbravo/event/PaidStatusEventHandler.java
+++ b/src/org/openbravo/event/PaidStatusEventHandler.java
@@ -73,7 +73,7 @@ class PaidStatusEventHandler extends EntityPersistenceEventObserver {
                     && (!psd.getPaymentDetails().isPrepayment())) {
                   psd.setInvoicePaid(true);
                 }
-                if (psd.isInvoicePaid()) {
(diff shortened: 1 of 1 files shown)


src/org/openbravo/event/PaidStatusEventHandler.java
etendobot, line 1: # GPT Review for PaidStatusEventHandler.java
The code in the git diff looks fine, but there is one improvement that can be made. Instead of calling `psd.isInvoicePaid()` twice, we can store the result in a variable and use that variable in the if statement. This will improve code readability and avoid unnecessary method calls.

Here's the updated code:

```java
if (psd.isInvoicePaid()) {
  boolean isReconciliationNull = transaction.getReconciliation() == null;
  if (isReconciliationNull) {
    FIN_Utility.updatePaymentAmounts(psd);
    FIN_Utility.updateBusinessPartnerCredit(transaction.getFinPayment());
  }
}
```

By storing the result of `psd.isInvoicePaid()` in the `isReconciliationNull` variable, we avoid calling the method twice and make the code easier to read.

Issue #224, opened 2023-09-08
EPL-895:Error adding line in sales order


brunoromerosmf:
**Error’s description**
When adding a line in a sales order, when a product is selected, the related callout is broken (output of when it happens).

2023-09-07 10:22:13,074 [http-2] ERROR org.openbravo.client.application.window.FormInitializationComponent - TabId:187 - Mode:CHANGE - rowId:_1694074928218
org.openbravo.base.exception.OBException: Couldn't execute callout (class org.openbravo.erpCommon.ad_callouts.SL_Order_Product)
 at org.openbravo.client.application.window.FormInitializationComponent.runCallouts(FormInitializationComponent.java:1518) ~[classes/:?]
 at org.openbravo.client.application.window.FormInitializationComponent.executeCallouts(FormInitializationComponent.java:1434) ~[classes/:?]
 at org.openbravo.client.application.window.FormInitializationComponent.execute(FormInitializationComponent.java:271) [classes/:?]

This is because the product in the price list related to the sales order has the value 0. This causes an arithmetic exception to be thrown and the callout does not work correctly.

The exception is reproduced exactly in line 122 of class SL_Order_Product.

**Steps to reproduce the error**
.Create a sales order with example values (no influence).

.Create a new line in the sales order.

.Select a product that has 0 values in the price list of the sales order (edit the price list if necessary).

**Expected behavior**
You should get an error message explaining the error (or a warning) but the callout should still work correctly.

**Affected Version**
last version

PR #225, opened 2023-09-12
Issue #224: Error adding line in sales order

jortolanosmf:
It was detected that the problem in the SL_Order_Product callout occurs due to a division by 0 in the discount calculation. This occurs when the price in the rate is 0.

Validation was added so that when this case occurs, now when the price in the rate is 0 the calculated discount will also be 0

### DIFF
--- a/src/org/openbravo/erpCommon/ad_callouts/SL_Order_Product.java
+++ b/src/org/openbravo/erpCommon/ad_callouts/SL_Order_Product.java
@@ -119,9 +119,10 @@ public class SL_Order_Product extends SimpleCallout {
     BigDecimal calculatedDiscount = BigDecimal.ZERO;
     BigDecimal price = isTaxIncludedPriceList ? grossPriceList : netPriceList;
     if (!BigDecimal.ZERO.equals(price)) {
+      int precision = order.getCurrency().getPricePrecision().intValue();
       calculatedDiscount = price.subtract(priceActual)
               .multiply(BigDecimal.valueOf(100))
-              .divide(price, 2)
+              .divide(price, precision)
               .setScale(0, RoundingMode.HALF_UP);
     }
     if (calculatedDiscount.compareTo(info.getBigDecimalParameter("inpdiscount")) != 0) {
--- a/src/org/openbravo/erpCommon/ad_callouts/SL_Order_Product.java
+++ b/src/org/openbravo/erpCommon/ad_callouts/SL_Order_Product.java
@@ -116,11 +116,14 @@ public class SL_Order_Product extends SimpleCallout {
     }
 
     // Discount
+    BigDecimal calculatedDiscount = BigDecimal.ZERO;
     BigDecimal price = isTaxIncludedPriceList ? grossPriceList : netPriceList;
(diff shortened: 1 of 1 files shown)


src/org/openbravo/erpCommon/ad_callouts/SL_Order_Product.java
etendobot, line 1: # GPT Review for SL_Order_Product.java
The code provided in the git diff looks fine, but there are a few improvements that can be made:

1. Move the declaration of `calculatedDiscount` outside of the `if` statement. This will make the variable accessible outside of the `if` block and avoid duplicate declarations.
2. Use the `compareTo(BigDecimal.ZERO)` method instead of `!BigDecimal.ZERO.equals(price)` to check if `price` is not equal to zero.
3. Use the `setScale(0, RoundingMode.HALF_UP)` method directly on the result of the division operation, instead of on the final value of `calculatedDiscount`.

Here's the updated code:

```java
protected void execute(CalloutInfo info) throws ServletException {
  // ...

  // Discount
  BigDecimal calculatedDiscount = BigDecimal.ZERO;
  BigDecimal price = isTaxIncludedPriceList ? grossPriceList : netPriceList;
  if (price.compareTo(BigDecimal.ZERO) != 0) {
    calculatedDiscount = price.subtract(priceActual)
        .multiply(BigDecimal.valueOf(100))
        .divide(price, 2, RoundingMode.HALF_UP);
  }
  if (calculatedDiscount.compareTo(info.getBigDecimalParameter("inpdiscount")) != 0) {
    info.addResult("inpdiscount", calculatedDiscount);
  }
}
```

This code improves readability and removes unnecessary duplicate code.

Gremiger (approved):

Matias-Bernal (approved):

src/org/openbravo/erpCommon/ad_callouts/SL_Order_Product.java
etendobot, line 1: # GPT Review for SL_Order_Product.java
Based on the provided git diff, here are the improvements that can be made to optimize the code:

1. Move the calculation of the `precision` variable outside the if condition since it is not dependent on the condition.
2. Use the `setScale` method with the `RoundingMode.HALF_UP` argument directly on the `calculatedDiscount` variable instead of using `divide` and then `setScale`.

Here's the modified code:

```java
protected void execute(CalloutInfo info) throws ServletException {
    BigDecimal calculatedDiscount = BigDecimal.ZERO;
    BigDecimal price = isTaxIncludedPriceList ? grossPriceList : netPriceList;
    int precision = order.getCurrency().getPricePrecision().intValue();
    if (!BigDecimal.ZERO.equals(price)) {
        calculatedDiscount = price.subtract(priceActual)
                .multiply(BigDecimal.valueOf(100))
                .divide(price, precision, RoundingMode.HALF_UP);
    }
    if (calculatedDiscount.compareTo(info.getBigDecimalParameter("inpdiscount")) != 0) {
        // Rest of the code
    }
}
```

By moving the calculation of `precision` outside the if condition, we avoid unnecessary calculations in case the condition is not met. Additionally, using `setScale` directly on `calculatedDiscount` simplifies the code and avoids an extra method call.

PR #226, opened 2023-09-25
Feature ETO-248: Update Wiki URLs

valeg-etendo:


build.gradle
etendobot, line 1: # GPT Review for build.gradle
Based on the git diff, it seems like the code is updating some XML content in a file. The change is replacing a URL in the `adModuleContent` variable.

The improvement I would suggest is to use a more specific and reliable regular expression pattern to match the URL. The current pattern is too generic and may match unintended parts of the content. Here's an example of an improved regular expression pattern:

```groovy
adModuleContent = adModuleContent.replaceAll(/<UPDATEINFO><!\[CDATA\[https:\/\/docs\.etendo\.software\/whats-new\/release-notes\/etendo-classic\/versions\/\d+Q\d+-\d+/, "<UPDATEINFO><![CDATA[https://docs.etendo.software/whats-new/release-notes/etendo-classic/versions/${version[0]}Q${version[1]}-${version[2]}")
```

In this pattern, I've escaped the dots in the URL to match them literally, and added a backslash before each dot. This ensures that the dots in the URL are treated as literal characters and not as regular expression metacharacters.

Additionally, it's a good practice to use a constant or a configuration value for the URL instead of hardcoding it in the regular expression. This makes the code more maintainable and allows for easier changes in the future.

etendobot, line 1: # GPT Review for build.gradle
The code in the git diff seems to be replacing some values in a string. Here are some improvements that can be made:

1. Use a more descriptive variable name instead of `adModuleContent`. This will make the code more readable and easier to understand.

2. Instead of using multiple `replaceAll` calls, you can use a single regular expression with capturing groups to replace all the values at once. This will improve performance and simplify the code.

3. Extract the URL pattern into a separate variable to make it easier to modify in the future.

Here's an example of how the code can be improved:

```groovy
task upgradeCoreVersion {
    doLast {
        String filePath = 'path/to/file.xml'
        String xmlContent = file(filePath).text

        String versionPattern = /(\d+)Q(\d+)-(\d+)/
        String urlPattern = /https:\/\/docs.etendo.software\/whats-new\/release-notes\/etendo-classic\/versions\/${versionPattern}/

        String updatedContent = xmlContent.replaceAll(versionPattern) { match ->
            def version = match[1..3]
            "<VERSION><![CDATA[${version.join('.')}]]>"
        }.replaceAll(urlPattern) { match ->
            def version = match[1..3]
            "<UPDATEINFO><![CDATA[${urlPattern.replace(versionPattern, version.join('Q'))}]]>"
        }

        file(filePath).write(updatedContent)
    }
}
```

In this example, I've used capturing groups in the regular expressions to extract the version numbers. Then, I've used a closure to generate the replacement string based on the captured groups. This approach allows for more flexibility and easier modification in the future.

Issue #260, opened 2023-11-30
EPL-571: Error is not displayed correctly when deleting a business partner and throwing an exception


valenvivaldi:
**Error's description**
When a business partner is deleted and an exception is triggered, the message appears only for a few seconds and then is deleted making it impossible to read the message.
This error occurs both in demo and livebuilds of openbravo, so I was doing debug and I suspect that the problem happens in smart client, as it does not happen with other entities, I tested it with products, invoices, and payments.

**Steps to reproduce the error**
try to delete any business partner, you will see an error that this business partner is related to other objects and the error will disappear.

PR #261, opened 2023-11-30
Issue #260: prevent hide message bar when refresh child tab

valenvivaldi:
This error happens when the parent tab shows a message and a child that is from the same entity (like the Business Partner window) is refreshed. This fix prevents the hide of the message bar in the parent tab.


Gremiger (approved):

Gremiger:
The commit title should not surpass the 70 characters.

valenvivaldi:
> The commit title should not surpass the 70 characters.

Fixed

isaiasb-etendo (approved):

Issue #264, opened 2023-12-01
EPL-1146: Database crashes when executing 'delete client' process with large volume of data


brunoromerosmf:
# Error's description
 During testing of the resolved error https://github.com/etendosoftware/etendo_core/issues/227  in the Delete Client process, a problem was 
 encountered when attempting to delete entities with large volumes of data. This was tested in a client's backup environment, and it 
 resulted in database locks that caused the delete process to never complete.

# Steps to reproduce the error
  1. Attempt to delete entities with large volumes of data.
  2. Observe the database locks and the failure of the delete process to complete.
  
# Expected behavior
  The Delete Client process should handle entities with large volumes of data without causing database locks, and the process should complete successfully.

# Affected Version
  Etendo Core 23.3.2

PR #275, opened 2023-12-27
Issue #264: Added deleted client gradle task process.

Matias-Bernal:
EPL-1146: Also fixed issues with source format modules with dependencies to JAR format modules unable to find their tables.


src/org/openbravo/service/system/SystemService.java
```diff
@@ -467,6 +468,8 @@ private void enableConstraints(Platform platform) {
       platform.enableAllTriggers(con, xmlModel, false);
       log4j.info("   Enabling foreign keys");
       platform.enableAllFK(con, xmlModel, false);
+    } catch (Exception e) {
+      e.printStackTrace();
```
Gremiger, line 472: It's ok to only print the stack trace? We are not rethrowing this exception nor handling it?

Gremiger (approved):

lucas-palacios (approved):

Issue #282, opened 2024-01-08
EPL-1197: Problem when creating a "payment in" or "payment out" from the creation of a transaction line in the financial account window.


brunoromerosmf:
# Error's description

Upon completing the creation of a transaction line in the 'Financial Account' window, where we create it without using the form mode, and where we proceed to generate either a 'payment in' or a 'payment out', we encounter an issue. While the 'payment in' or 'payment out' is successfully generated (and can subsequently be viewed in the respective windows), the transaction line itself is not created and disappears. Furthermore, refreshing the tab results in an error, as depicted in the attached document.

![image](https://github.com/etendosoftware/etendo_core/assets/122372040/6c675367-fbda-41dd-9963-b5acaa9a1931)

# Steps to reproduce the error

1. Go to the "account financial" window
2. Create a transaction line from the grid, and in it, create a payment in/ payment out.
3. save, you will see that the log disappears, then refresh. And there you will see the error.

# Expected behavior

The transaction should be created correctly

# Affected version

23.4.0

PR #300, opened 2024-01-26
Issue #282: Change the validation to check for undefined too

RomanMagnoli:
EPL-1197.

(26/01) - With the implementation of Actions, now the Object returned from the AddPaymentActionHandler is an ActionResult. This causes the properties added as "false" are not included in the response, it just avoids adding at least they were "true".
Arriving at the JS (modules_core/org.openbravo.client.application/web/org.openbravo.client.application/js/process/ob-parameter-window-view.js) this response (data) doesn't have the properties needed to be checked, they are "undefined" and comparing "undefined === false" is not the same that "false === false", so I have changed the validation to check for undefined too.

(29/01) - Change the logic of the validation to avoid being confusing.

### DIFF
--- a/modules_core/org.openbravo.client.application/web/org.openbravo.client.application/js/process/ob-parameter-window-view.js
+++ b/modules_core/org.openbravo.client.application/web/org.openbravo.client.application/js/process/ob-parameter-window-view.js
@@ -484,7 +484,7 @@ isc.OBParameterWindowView.addProperties({
         },
         function(rpcResponse, data, rpcRequest) {
           view.handleResponse(
-            !(data && data.refreshParent !== true),
+            data && data.refreshParent === true),
             data && data.message,
             data && data.responseActions,
             data && data.retryExecution,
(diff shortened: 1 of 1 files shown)


valenvivaldi (approved):

Matias-Bernal (approved):

modules_core/org.openbravo.client.application/web/org.openbravo.client.application/js/process/ob-parameter-window-view.js
```diff
@@ -484,7 +484,7 @@ isc.OBParameterWindowView.addProperties({
         },
         function(rpcResponse, data, rpcRequest) {
           view.handleResponse(
-            !(data && data.refreshParent === false),
+            !(data && !data.refreshParent),
```
sebastianbarrozo, line 487: The double negation is a bit confusing, subject to testing, isn't it equivalent to: "data && data.refreshParent !== true"?

sebastianbarrozo (changes requested):

modules_core/org.openbravo.client.application/web/org.openbravo.client.application/js/process/ob-parameter-window-view.js
RomanMagnoli: Yes, it's equivalent. I have just changed the logic because "undefined === false" is not the same that "false === false". I'll change it to data.refreshParent !== true. It should be the same

Gremiger (approved):

valenvivaldi (approved):

valeg-etendo (approved):

Matias-Bernal (approved):

Matias-Bernal (approved):

sebastianbarrozo (approved):

Issue #290, opened 2024-01-22
EPL-1214: Description field size


juanfunes1996:
### **Error's description**
The description fields are now limited to a single line. No matter how large you make them in the view (in window editing), they still appear on a single line.

### **Steps to reproduce the error**
1 -Click to change the layout of the fields in a window like the sales order window.
2 - Edit the number of rows in the description field.
![image](https://github.com/etendosoftware/etendo_core/assets/143511686/07c775e8-e361-476e-b9e9-2b3defd89dd0)

### **Expected behavior**
To be able to modify the field to have more lines

### **Affected Version**
23.4.0

### **Solution Design (optional)**
This is because in the skin change from OB to Etendo the field reference type was changed. From "text" to "String". 

![image](https://github.com/etendosoftware/etendo_core/assets/143511686/441a33c9-d723-458d-9bae-8489deaa739a)

PR #303, opened 2024-01-31
Issue #290: Description field size is changed from 'String' to 'Text' in C_Order table

tomyrodeghiero:
EPL-1214: The field reference type in C_Order table change from 'String' to 'Text'


Gremiger:
Is this PR ok? Because the description doesn't match with the entirety of the changelist

tomyrodeghiero:
> Is this PR ok? Because the description doesn't match with the entirety of the changelist

It is not okay. When I executed **export.database**, it brought in many changes that shouldn't have been impacted. I will soon commit to removing the unnecessary changes because I wasn't aware that there were changes that shouldn't be uploaded. I thought these were the correct ones. Thank you for letting me know.

valenvivaldi:
Always remember to check carefully what is added to the git stage before committing.

tomyrodeghiero:
> Always remember to check carefully what is added to the git stage before committing.

@valenvivaldi  I completely agree. I thought the changes I was uploading (since it was the only thing I modified in **export.database**, and I had a clean tree of changes before running it) were fine. Thank you very much for the advice, I agree that it's essential to do it before committing

@isaiasb-etendo will soon provide me with a command to clean up the things that shouldn't have been uploaded

isaiasb-etendo (approved):

valenvivaldi (approved):

Gremiger (approved):

Issue #308, opened 2024-02-06
EPL-1266: Error in "Create invoice from orders" window according to invoice type


juanfunes1996:
# Error's description

A fully delivered order cannot be invoiced when selecting 'After order Delivered” in the billing type selector on the sales order.

# Steps to reproduce the error

1. Create a Sales Order in the 'Sales Order' window, select 'After Order Delivered' in the 'Invoice Terms' dropdown menu.
2. Create a line for that order with a product.
3. Complete the order.
4. In the 'Goods Shipment' window, create a header.
5. Select the sales order and the line created previously through the 'Create Lines From' button and click 'Ok.'
6. Complete record.
7. Go to the 'Create Invoices from Order' window and we will not see the created order.

https://github.com/etendosoftware/etendo_core/assets/143511686/2af6fa00-261c-4837-933a-3a4b97f334bf

# Expected behavior

The invoice will be generated after all products from the sales order have been shipped.

# Affected Version

Core : 23.4.1

valeg-etendo:
> [!IMPORTANT]
> Bug resolved in version [23.4.3](https://github.com/etendosoftware/etendo_core/releases/tag/23.4.3).

PR #324, opened 2024-02-28
Issue #308: Fixed 'Generate invoices from orders' process

faqsarg:
EPL-1266: The process was missing the case for 'After Order Delivered' invoice term. Also improved query identation.


faqsarg:
sorry. fixed identation so it's easier to review.

Issue #343, opened 2024-03-21
[EPL-1414] Standard process definition add lines on a tab but you have to refresh to see them reflected


brunoromerosmf:
# Error's description
In a standard process definition, when new lines are added in relation to a header, the database is updated accordingly. However, to see these updates reflected on the interface, it is necessary to refresh the window. To be more precise, all the classes (processes) that extend BaseProcessActionHandler.

 # Steps to reproduce the error
The example is with the sales order window:
1. Create a sales order
2. Click in “Copy from orders” button
3. Select as many lines as you want.
4. Done

 # Expected behavior
It should not be necessary to refresh the window to see the new lines reflected.
 # Affected Version
¿ Latest/Specific ?

Starting from: 23.4.1

To: 24.1.1

# Solution design (optional)

Use as a solution concept the one applied by Matias in the issue attached below. 
https://github.com/etendosoftware/etendo_core/pull/337

# Other test cases
  Given: in sales order
  When: you run the buton of process "copy from orders" and select x number of lines
  Then: In the "lines" tab the selected lines should be loaded automatically without refreshing. 

 

  Given: in sales invoices
  When: you run the buton of process "create lines from order" and select x number of lines 
  Then: In the "lines" tab the selected lines should be loaded automatically without refreshing.

 

  Given: in sales invoices
  When: you run the buton of process "create lines from shipment" and select x number of lines 
  Then: In the "lines" tab the selected lines should be loaded automatically without refreshing.

 

  Given: in purchase order
  When: you run the buton of process "copy from orders" and select x number of lines 
  Then: In the "lines" tab the selected lines should be loaded automatically without refreshing.

 

  Given: in purchase order
  When: you run the buton of process "create lines" and select x number of lines
  Then: In the "lines" tab the selected lines should be loaded automatically without refreshing.

 

  Given: in purchase invoice 
  When: you run the buton of process "create lines from receipt" and select x number of lines
  Then: In the "lines" tab the selected lines should be loaded automatically without refreshing.

 

  Given: in purchase invoice
  When: you run the buton of process "create lines from order" and select x number of lines
  Then: In the "lines" tab the selected lines should be loaded automatically without refreshing.

valeg-etendo:
> [!IMPORTANT]
> Bug resolved in version [24.1.1](https://github.com/etendosoftware/etendo_core/releases/tag/24.1.1).

valeg-etendo:
> [!IMPORTANT]
[...]

PR #357, opened 2024-04-10
Issue #343: Add 'refreshParent' property to the BaseProcessActionHandler process

RomanMagnoli:
EPL-1414


modules_core/org.openbravo.client.application/src/org/openbravo/client/application/process/BaseProcessActionHandler.java
```diff
@@ -167,6 +169,12 @@ protected final JSONObject execute(Map<String, Object> parameters, String conten
     }
   }
 
+  private JSONObject doRefreshParent(JSONObject result) throws JSONException {
+    boolean deRefresh = !result.has(REFRESH_PARENT) || result.getBoolean(REFRESH_PARENT);
```
Gremiger, line 173: result.has(REFRESH_PARENT) && result.getBoolean(REFRESH_PARENT)   ?

Matias-Bernal (approved):

modules_core/org.openbravo.client.application/src/org/openbravo/client/application/process/BaseProcessActionHandler.java
RomanMagnoli: The main idea is to have the refresh on true as default if it's not set. This is:
 - If the check existed, I need to set it as it came. So, the first check is to return true if it doesn't exist. As it's an 'or' and Java has lazy validations, the first true will return.
 - The second part of the 'or', is, in case the refresh exists, so if the refresh existed I need to use the boolean that was set, so I "false" the first condition to valid the boolean with the second one

Gremiger (approved):

valeg-etendo (approved):

modules_core/org.openbravo.client.application/src/org/openbravo/client/application/process/BaseProcessActionHandler.java
etendobot, line 1: # GPT Review for BaseProcessActionHandler.java
### Review

- **Estimated effort to review [1-5]:**  2, because the changes are focused and relatively easy to understand.
- **Score:**  95

### Code feedback

- **File:** 
modules_core/org.openbravo.client.application/src/org/openbravo/client/application/process/BaseProcessActionHandler.java
- **Language:** 
Java
- **Suggestion:** 
Consider adding null checks to handle potential `NullPointerException` in the `doRefreshParent` method [important]
- **Label:** 
bug
- **Existing code:**

```java
private JSONObject doRefreshParent(JSONObject result) throws JSONException {
    boolean deRefresh = !result.has(REFRESH_PARENT) || result.getBoolean(REFRESH_PARENT);
    result.put(REFRESH_PARENT, deRefresh);
    return result;
}
```
- **Improved code:**

```java
private JSONObject doRefreshParent(JSONObject result) throws JSONException {
    if (result == null) {
        throw new IllegalArgumentException("Result object cannot be null");
    }
    boolean deRefresh = !result.has(REFRESH_PARENT) || result.getBoolean(REFRESH_PARENT);
    result.put(REFRESH_PARENT, deRefresh);
    return result;
}
```

src/org/openbravo/common/actionhandler/CopyFromOrdersActionHandler.java
etendobot, line 1: # GPT Review for CopyFromOrdersActionHandler.java
This file has no corrections to be made. Well done!

PR #515, opened 2024-10-28
Feature ETP-43: Epic of product validation

valenvivaldi:


valenvivaldi:
@sebastianbarrozo Seba, here we have some changes that I have in my local environment and I understand that this is in the product validation server, this should reach Develop de Etendo Core, no?

PR #827, opened 2025-11-07
Feature ETP-2832: Upgrade arquillian version

sebastianbarrozo:


modules_core/org.openbravo.service.json/src/org/openbravo/service/json/DefaultJsonDataService.java
```diff
@@ -93,10 +93,13 @@ public class DefaultJsonDataService implements JsonDataService {
   @Inject
   private CachedPreference cachedPreference;
 
-  private static DefaultJsonDataService instance = WeldUtils
-      .getInstanceFromStaticBeanManager(DefaultJsonDataService.class);
+  private static DefaultJsonDataService instance = null;
```
etendobot: **Suggestion**

Consider declaring the 'instance' variable as 'volatile' to ensure thread safety for lazy initialization in multi-threaded scenarios.

```java
private static volatile DefaultJsonDataService instance = null;
```

**Rationale**: While the getInstance() method ensures the instance is only created on first use, without 'volatile' there is a slight risk of seeing a partially constructed object in rare thread contention cases (see Java Memory Model best practices on double-checked locking). Marking the instance as volatile provides an extra guarantee and is a common defensive measure. This is a suggestion and not required by Etendo's default patterns, but aligns with modern Java singleton best practices.

src-test/src/org/openbravo/advpaymentmngt/test/FundsTransferTest.java
```diff
@@ -123,7 +123,7 @@ private void createConversionRate(Currency from, Currency to) {
   }
 
   @Test
-  @InSequence(1)
+  @Order(1)
   public void testFundsTransferSameCurrency() {
```
etendobot: **Suggestion**

Review that the class uses @TestMethodOrder(OrderAnnotation.class) to guarantee @Order annotations work as expected for JUnit 5. While this may already exist due to project conventions, explicit inclusion prevents order issues.

```java
@Order(1) // Ensure @TestMethodOrder(OrderAnnotation.class) is set at class level if not present
```

**Rationale**: JUnit 5 requires annotating the test class with @TestMethodOrder(OrderAnnotation.class) for @Order to be effective. Code search and Etendo documentation on test structure recommend this as best practice during migration.

etendobot (approved):

PR #879, opened 2026-01-16
Feature ETP-3155: Fixed webservices test

Matias-Bernal:
This pull request makes a small change to the `BaseDataSourceTestDal` test class, converting the `authenticated`, `cookie`, and `csrfToken` fields from static to instance variables. This change improves test isolation by ensuring these fields are not shared across test instances.


src-test/src/org/openbravo/test/datasource/BaseDataSourceTestDal.java
```diff
@@ -45,9 +45,9 @@ public class BaseDataSourceTestDal extends OBBaseTest {
   private static String OB_URL = null;
   protected static final String LOGIN = "admin";
   protected static final String PWD = "admin";
-  private static boolean authenticated = false;
-  private static String cookie;
-  private static String csrfToken;
+  private boolean authenticated = false;
+  private String cookie;
```
etendobot: **Suggestion**

Verify if any existing tests relying on this base class expected a shared (static) authentication state.

```java
private boolean authenticated = false;
```

**Rationale**: Changing 'authenticated', 'cookie', and 'csrfToken' from static to instance variables ensures that authentication state is not shared between different test instances. This prevents side effects where one test's authentication might affect another, especially in parallel execution or complex test suites. Since this is a test base class, this change is a common practice to ensure test isolation.

pipelines/unittests/JenkinsfileOracle
```diff
@@ -176,9 +162,9 @@ pipeline {
 
                             def stackConfiguration = stackConfig.resolveStackConfiguration(env.FROM_BACKPORT, env.FROM_PRERELEASE)
 
-                            env.JAVA_HOME = stackConfiguration.javaHome
-                            env.TOMCAT_URL = stackConfiguration.tomcatUrl
-                            env.STACK_TYPE = stackConfiguration.stackType
+                            env.JAVA_HOME = "/usr/lib/jvm/jdk-17.0.13"
+                            env.TOMCAT_URL = "https://dlcdn.apache.org/tomcat/tomcat-11/v11.0.15/bin/apache-tomcat-11.0.15.tar.gz"
```
etendobot: > [!WARNING]
**⚠️ Blocking Issue**

Revert to using the dynamic 'stackConfiguration.javaHome' instead of a hardcoded path.

```groovy
env.JAVA_HOME = stackConfiguration.javaHome
```

**Rationale**: Hardcoding paths to specific JDK versions makes the pipeline fragile and dependent on the specific configuration of the Jenkins agent. It overrides the dynamic resolution logic designed to handle multiple environments.

pipelines/unittests/JenkinsfileOracle
```diff
@@ -253,8 +239,15 @@ pipeline {
                                 mkdir -p build/lib/runtime
                                 ./gradlew prepareConfig --info
                                 ./gradlew setup --info
-                                ./gradlew install --info
                                 ./gradlew expandModules --info
```
etendobot: > [!WARNING]
**⚠️ Blocking Issue**

Ensure 'install' is executed in the correct sequence relative to 'expandModules' or explain why the order was swapped.

```groovy
./gradlew setup --info
./gradlew install --info
```

**Rationale**: The change splits the Gradle command block and moves 'install' after the database wait. However, 'expandModules' is now executed before 'install'. In Etendo, the install task typically handles database schema creation which is often a prerequisite for module expansion or proper state in complex environments. Moving 'install' after 'expandModules' without clear justification may lead to inconsistent build states.

Issue #887, opened 2026-01-27
ETP-3230: Unpost/Reactivate allows modifying accounting entries in closed periods for invoices with accruals


juanfunes1996:
# Error's description
When a vendor invoice with expense accruals (for example, a 12-month distribution) has already generated accounting entries across multiple periods, the system allows the user to Unpost and Reactivate the document even when some or all of the affected periods are closed.

This behavior is inconsistent with standard period control validations, since the system correctly blocks the posting of new documents in closed periods by displaying the message:

> "The Period does not exist or it is not opened"

However, this validation is not applied to the unpost/reactivate process, allowing accounting entries belonging to closed periods to be deleted or modified, which compromises accounting integrity and audit traceability.

 # Steps to reproduce the error
1. Ensure the product used in the invoice line is properly configured for accruals.
2. Verify in the Accounting tab that the [Product Deferred Expense](https://docs.etendo.software/user-guide/etendo-classic/how-to-guides/how-to-manage-deferred-revenue-and-expenses/#execution-steps-expenses) account is defined.This is the account used to post deferred or prepaid expenses for purchases of this product. It is inherited from the product category and can be further defined at product level.
3. Create a vendor invoice with a 12-month expense accrual starting on 01/01/2025.
4. Complete and post the invoice so that accounting entries are generated for all months of the year.
5. Close the periods from January to November, leaving only December open.
6. Execute Unpost on the invoice.
7. Observe that the system allows the action without showing any error.

Additional validation:
Try to post a new invoice dated in a closed period → the system correctly blocks the action with the message:
"The Period does not exist or it is not opened"

https://github.com/user-attachments/assets/614936b3-0687-42e7-8e01-f1ab241b485a

 # Expected behavior
The system should block unposting or reactivating any document if this implies modifying or deleting accounting entries belonging to closed periods.
At a minimum, one of the following validations should be enforced:

- Block the action if any Fact_Acct record related to the document belongs to a closed period.
- Or require reopening the affected periods before allowing the action.

This ensures accounting integrity, traceability, and audit compliance.

 # Affected Version

- Etendo Classic 25.4.3
- Etendo Classic 24.4.31

 # Solution Design (optional)

 # Other test cases
 **Given:**
 **When:**
 **Then:**

PR #896, opened 2026-02-02
Issue #887: Fix unpost allowed in closed periods

Agustin-Calderon:
ETP-3230-Y24

### DIFF
--- a/src/org/openbravo/financial/ResetAccounting.java
+++ b/src/org/openbravo/financial/ResetAccounting.java
@@ -19,7 +19,6 @@
 package org.openbravo.financial;
 
 
-import java.sql.SQLException;
 import java.text.ParseException;
 import java.util.ArrayList;
 import java.util.Arrays;
@@ -38,11 +37,14 @@ import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 import org.hibernate.ScrollMode;
 import org.hibernate.ScrollableResults;
+import org.hibernate.criterion.DetachedCriteria;
+import org.hibernate.criterion.Projections;
+import org.hibernate.criterion.Property;
 import org.hibernate.criterion.Restrictions;
+import org.hibernate.criterion.Subqueries;
 import org.hibernate.query.Query;
 import org.openbravo.base.exception.OBException;
 import org.openbravo.base.model.ModelProvider;
-import org.openbravo.client.application.process.ResponseActionsBuilder;
 import org.openbravo.dal.core.OBContext;
 import org.openbravo.dal.security.OrganizationStructureProvider;
 import org.openbravo.dal.service.OBCriteria;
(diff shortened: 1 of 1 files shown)


src/org/openbravo/financial/ResetAccounting.java
```diff
@@ -193,6 +196,13 @@ public static HashMap<String, Integer> delete(final String adClientId, final Str
               scroll.close();
             }
 
+            if (StringUtils.isNotBlank(localRecordId)) {
+              Set<String> orgPeriodControls = new HashSet<>(organizationPeriodControl.values());
```
etendobot: > [!WARNING]
**⚠️ Blocking Issue**

Refactor the validation to check all records in the batch at once or integrate the period status check into the main scrollable query.

```java
// It is recommended to perform a bulk validation outside the loop or 
// use a single query that joins the main record selection with period control.
```

**Rationale**: Executing a validation query (validateNoFactsInClosedPeriods) inside the loop for every 'localRecordId' will lead to N+1 query issues. For large datasets, this will significantly slow down the Reset Accounting process.

src/org/openbravo/financial/ResetAccounting.java
```diff
@@ -311,11 +321,70 @@ tableId, orgIds, clientId, p[0], p[1], getCalendarId(organization), strdatefrom,
...
+        .forEntityName("FinancialMgmtPeriodControl", "pc")
+        .setProjection(Projections.property("pc.id"))
+        .createAlias("pc.period", "pcperiod")
+        .add(Property.forName("pcperiod.id").eqProperty("faperiod.id"))
+        .add(Restrictions.eq("pc.organization.id", orgPeriodControl))
+        .add(Restrictions.eq("pc.documentCategory", docBaseType))
+        .add(Restrictions.eq("pc.periodStatus", "O"));
+
+    factCrit.add(Subqueries.notExists(pcExists));
+
```
etendobot: **Suggestion**

While the logic is technically correct for identifying closed periods, the implementation inside the loop is the primary concern.

```java
    factCrit.add(Subqueries.notExists(pcExists));
    // Ensure this validation only triggers if there are actually facts to be deleted.
    // If the list is empty because no facts exist, it shouldn't throw an exception.
    // However, the current logic throws an exception if a fact exists in a period that is NOT open.
    // Ensure that if a record has NO facts, it doesn't fail.
```

**Rationale**: The use of 'notExists' in the criteria means that if an Accounting Fact exists but its period is NOT 'O' (Open), the query returns results and the exception is thrown. This is correct for closed periods. However, if no facts exist for the record (e.g., it was never posted), 'factCrit.list()' will be empty and no exception is thrown, which is correct. The blocking issue remains the performance impact of the loop.

etendobot (changes requested):

Issue #903, opened 2026-02-10
ETP-3355: Incorrect error handling in Tree View: OBException messages not displayed to the user


marciomartinelli22:
# Error's description

When an OBException is thrown from an Event Handler during an operation in the Tree View (e.g., Account Tree), the error message is not correctly displayed to the user. The frontend expects the type field in the JSON error response, but the backend (JsonUtils.java) serializes the error type into the messageType field. This causes the Tree View to not recognize the error and fail to render it properly.

 # Steps to reproduce the error

Required Configurations

A module implementing an Event Handler on any entity that uses a Tree View (e.g., C_ElementValue) which throws an OBException with a custom message.

Create an Event Handler that validates a condition on a Tree View entity and throws an OBException if the validation fails.
Navigate to the corresponding Tree View window (e.g., Account Tree).
Perform an operation that triggers the exception (e.g., moving a node or saving a record that fails the validation).
Observe that the custom error message is not displayed to the user, or a generic/malformed error appears instead.

Root cause:

In JsonUtils.java, the error type is placed in the messageType field when building the JSON error response.
In the frontend (ob-tree-view-grid.js, ob-standard-view.js, ob-utilities.js), the code evaluates error.type or response.error.type, never error.messageType.
A fallback exists in ob-standard-view.js:563 (type = error.messageType || type), but in the Tree View code path this fallback does not apply, so the error is never correctly classified and not rendered to the user.
 
# Expected behavior

When an Event Handler throws an OBException during an operation in the Tree View, the user should see a clear error message containing the exception text, in the same way errors are displayed in standard form/grid views.

 # Affected Version

Etendo 25

 # Solution Design (optional)

Fix the inconsistency in the core so that the error type field is consistent between backend and frontend:
Backend (JsonUtils.java):

When serializing errors, use the type field instead of (or in addition to) messageType, to match what the frontend expects.

Frontend (ob-tree-view-grid.js):

Ensure the Tree View error handler correctly reads the type field from the error response and displays the message to the user.

Justification:

All frontend JavaScript code already evaluates error.type (ob-tree-view-grid.js, ob-standard-view.js, ob-utilities.js).
messageType is inconsistent and causes errors to not be processed correctly in the Tree View.
The existing fallback in ob-standard-view.js (error.messageType || type) guarantees backward compatibility.
This fix benefits all modules using Event Handlers with Tree Views, not just a specific case.

 # Other test cases

[...]

PR #904, opened 2026-02-11
Issue #903: View OBException in accounting tree view

marciomartinelli22:
ETP-3355

### DIFF
--- a/modules_core/org.openbravo.service.json/src/org/openbravo/service/json/JsonUtils.java
+++ b/modules_core/org.openbravo.service.json/src/org/openbravo/service/json/JsonUtils.java
@@ -247,13 +247,13 @@ public class JsonUtils {
         } else {
           error.put("message", "OBUIAPP_QueryTimeOut");
         }
-        error.put("messageType", obError.getType());
+        error.put("type", obError.getType());
         error.put("title", obError.getTitle());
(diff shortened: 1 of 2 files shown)


modules_core/org.openbravo.service.json/src/org/openbravo/service/json/JsonUtils.java
```diff
@@ -247,13 +247,13 @@ public static String convertExceptionToJson(Throwable throwable) {
         } else {
           error.put("message", "OBUIAPP_QueryTimeOut");
         }
-        error.put("messageType", obError.getType());
+        error.put("type", obError.getType());
         error.put("title", obError.getTitle());
```
etendobot: **Suggestion**

Consider keeping 'messageType' alongside 'type' to ensure backward compatibility with existing UI components that expect the legacy field.

```java
error.put("type", obError.getType());\n        error.put("messageType", obError.getType());
```

**Rationale**: The change replaces 'messageType' with 'type'. While 'type' is used in newer services (e.g., MessageService), historical JSON REST documentation and SmartClient integrations in Etendo expect 'messageType' within the error object. Removing 'messageType' might break legacy frontend components that rely on this specific key. Providing both ensures backward compatibility while supporting the new standard.

jortolanosmf (approved):

PR #918, opened 2026-02-13
test: add AccountTreeTest

sebastianbarrozo:


src-test/src/org/openbravo/erpCommon/businessUtility/AccountTreeTest.java
```diff
@@ -0,0 +1,1107 @@
...
+    AccountTree accountTree = createAccountTreeInstance();
+    setPrivateField(accountTree, "accountsTree", new AccountTreeData[0]);
+    setPrivateField(accountTree, "accountsFacts", new AccountTreeData[0]);
+    setPrivateField(accountTree, "reportElements", reportElements);
+    setPrivateField(accountTree, "reportNodes", new String[] {"0"});
+    return accountTree;
+  }
+
+  private AccountTree createAccountTreeInstance() throws Exception {
+    // Use reflection to create instance without calling constructor
```
etendobot: **Suggestion**

Consider if the class can be instantiated via its constructor by mocking the necessary parameters and internal static calls, rather than bypassing it with Objenesis.

```java
  private AccountTree createAccountTreeInstance() throws Exception { // Refactor to avoid Objenesis if possible
```

**Rationale**: Using Objenesis to bypass constructors can lead to fragile tests if the class relies on constructor-based initialization that isn't manually replicated via reflection. While acceptable for legacy code testing, it should be used sparingly.

etendobot (approved):

Issue #915, opened 2026-02-12
ETP-3371: Invoice Status in Goods Shipment is not calculated correctly on some cases


fran-roig:
# Error's description
The invoice status calculation in Goods Shipments has multiple issues:
1. When creating invoices and shipments in certain sequences, the second shipment is not created in Matched Sales Invoices and shows 0% invoiced status
2. In some sequences, it's not possible to create a third invoice because the shipment incorrectly shows 100% invoiced status
3. When voiding a shipment, the negative record is not created in Matched SO/SI

 # Steps to reproduce the error
Case 1: Missing Matched Sales Invoice Record
-   Create 1 Sales Invoice
-   Create 1 Goods Shipment
-   Create another Goods Shipment
-   Observe: The second shipment is not created in Matched Sales Invoices, invoiced status shows 0%

Case 2: Cannot Create Third Invoice

-   Create 1 Sales Invoice
-   Create 1 Goods Shipment
-   Try to create a second Sales Invoice
-   Observe: Cannot create the third invoice, the shipment incorrectly shows 100% invoiced

Case 3: Voided Shipment

-   Create a Sales Order with Goods Shipments
-   Void one of the shipments
-   Observe: No negative record is created in Matched SO/SI

 ### Required Configurations (if necessary)

Set to 'Y' the Active flag for Matched Sales Invoice and Matched Sales Order menues.

 # Expected behavior

- All shipments should be correctly registered in Matched Sales Invoices with accurate invoiced status percentages
- It should be possible to create invoices until the actual invoiced quantity matches the shipped quantity
- Voiding a shipment should create the corresponding negative record in Matched SO/SI to maintain data integrity

 # Affected Version
Etendo Core: Latest

 # Solution Design (optional)

 # Other test cases
 **Given:**
 **When:**
 **Then:**

valeg-etendo:
> [!IMPORTANT]
> Bug resolved in version [25.4.10](https://github.com/etendosoftware/etendo_core/releases/tag/25.4.10) ⚡ 🚀

PR #923, opened 2026-02-19
Issue #915: Fix MatchSI/SO for partial & void shipments

ivanrobledofutit:
ETP-3371

### DIFF
--- a/src-db/database/model/functions/C_INVOICE_POST.xml
+++ b/src-db/database/model/functions/C_INVOICE_POST.xml
@@ -1578,6 +1578,64 @@ IF (NOT FINISH_PROCESS) THEN
                AND C_INVOICELINE_ID = Cur_SILines.C_InvoiceLine_ID;
           END IF;
         END LOOP;
+
+        v_ResultStr := 'MatchSI-FromInvoice-OrderFallback';
+        FOR Cur_SILines IN
+          (SELECT il.AD_Client_ID,
+                  il.AD_Org_ID,
+                  il.C_InvoiceLine_ID,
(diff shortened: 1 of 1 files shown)


src-db/database/model/triggers/M_INOUTLINE_TRG.xml
```diff
@@ -82,7 +82,7 @@ BEGIN
      WHERE M_PRODUCT_ID = :NEW.M_PRODUCT_ID;
    END IF;
 
-   IF (v_QTYVARIABLE <> 'Y' AND v_ISSOTRX = 'Y' AND :NEW.C_ORDERLINE_ID IS NOT NULL) THEN
+   IF (v_QTYVARIABLE <> 'Y' AND :NEW.C_ORDERLINE_ID IS NOT NULL) THEN
        IF (UPDATING) THEN
```
etendobot: **Suggestion**

Ensure that the subsequent logic (UPDATING/INSERTING/DELETING) correctly handles the case for Purchase Orders (MMR) as well as Sales Orders (MMS).

```xml
IF (v_QTYVARIABLE <> 'Y' AND :NEW.C_ORDERLINE_ID IS NOT NULL) THEN
```

**Rationale**: Removing the 'v_ISSOTRX = 'Y'' condition extends the validation or update logic to both Sales Orders (Shipments) and Purchase Orders (Receipts). This is consistent with the PR title which mentions fixing MatchSI/SO (Sales/Purchase Invoice matching) for partial and void shipments. Extension to Purchase Orders is a valid business logic change.

Issue #926, opened 2026-02-23
ETP-3424 : Work Requirement can be processed even if Process Plan has no active version


juanfunes1996:
# Error's description
When a Work Requirement is created for a date in which the associated Process Plan has no active version, executing the Process Work Requirement action does not generate any phases, since there is no valid active version of the Process Plan.

However, despite no phases being generated and no actual processing taking place, the Work Requirement is still marked as Processed.

This results in an inconsistent state where the Work Requirement appears as processed even though the action could not be correctly executed.

 # Steps to reproduce the error
1. Configure a Process Plan with no active versions valid for day X.
2. Create a Work Requirement dated on day X, linked to that Process Plan.
3. Execute the Process Work Requirement action.

https://github.com/user-attachments/assets/cfab0b84-d899-44b4-8fd6-dfc214a04b92

 # Expected behavior
The system should display a clear and informative error message, for example:
“The Work Requirement cannot be processed because the selected Process Plan has no active version valid for the given date.”

The Work Requirement must not be marked as Processed if no phases could be generated.

 # Affected Version
Etendo Core: 25.4.7

<img width="1903" height="1011" alt="Image" src="https://github.com/user-attachments/assets/bb247f65-234d-483e-ab3a-6ff102a1397c" />

Etendo Core: 24.4.33

<img width="1903" height="1011" alt="Image" src="https://github.com/user-attachments/assets/cac49328-8129-4034-abb0-b1e53773ff87" />

 # Solution Design (optional)
Before marking the Work Requirement as processed, the system must validate that the associated Process Plan has at least one active version valid for the Work Requirement date.

 # Other test cases
 **Given:**
 **When:**
 **Then:**

PR #933, opened 2026-02-24
Issue #926: Add validation in Work Requirement processing when there is no active process plan version

ivanrobledofutit:
ETP-3424-Y24


src-db/database/model/functions/MA_WORKREQUIREMENT_PROCESS.xml
```diff
@@ -40,6 +40,7 @@
     v_ExplodePhases CHAR;
     v_Quantity NUMBER;
     v_Multiplier NUMBER;
+    v_CountVersion NUMBER;
     v_Processed CHAR;
```
etendobot: **Suggestion**

No changes needed here, just verifying the declaration aligns with the usage below.

```xml
v_CountVersion NUMBER;
```

**Rationale**: The new variable declaration is correct for the logic introduced.

etendobot (approved):

src-db/database/model/functions/MA_WORKREQUIREMENT_PROCESS.xml
```diff
@@ -116,6 +117,22 @@
     END IF; --FINISH_PROCESS
     IF(NOT FINISH_PROCESS) THEN
       IF(v_ExplodePhases='Y') THEN
+
+        SELECT COUNT(*)
```
Agustin-Calderon: Since we only need to validate existence, it would be better to use IF NOT EXISTS (SELECT 1 ...) instead of COUNT(*). This avoids scanning and counting all matching rows unnecessarily.

Agustin-Calderon (changes requested):

valeg-etendo (approved):

PR #940, opened 2026-02-24
Feature ETP-3433: Add new webservice to get AD_Messages translations

fran-roig:


src-test/src/com/etendoerp/common/rest/MessagesTrlWebServiceTest.java
```diff
@@ -0,0 +1,396 @@
...
+  // ---------------------------------------------------------------------------
+  // GET — happy path
+  // ---------------------------------------------------------------------------
+
+  /**
+   * GET with a valid {@code moduleId} and base language must return HTTP 200 and a non-empty
+   * JSON object with non-blank values.
+   */
+  @Test
+  public void getBaseLanguageReturnsMessages() throws Exception {
```
etendobot: **Suggestion**

Manually populate the JSONArray for better compatibility with Jettison JSON.

```java
  /** A POST with a missing {@code language} field must return HTTP 400. */
  @Test
  public void postMissingLanguageReturns400() throws Exception {
    JSONObject body = new JSONObject();
    JSONArray keys = new JSONArray();
    keys.put("Error");
    body.put(FIELD_SEARCH_KEYS, keys);
    doPostRequest(body.toString(), 400);
  }
```

**Rationale**: Depending on the specific version of Jettison used in the environment, the constructor JSONArray(Collection) might not behave exactly as expected or might be missing compared to standard org.json. Manual insertion is safer for compatibility.

PR #950, opened 2026-02-27
Feature ETP-3267: Fix org.openbravo.test.costing tests

RomanMagnoli:
This pull request introduces several improvements to dependency handling, fallback logic, and test isolation in the Openbravo core and test modules. The most significant changes include enhanced fallback mechanisms for bean instantiation, improved robustness in CDI/Weld utilities, and better test isolation and cleanup for JUnit tests. Additionally, costing tests have been updated to ensure correct preference handling and calculation accuracy.

Dependency handling and fallback improvements:

* Added fallback logic in `FIN_Utility.getInstance()` to use `OBProvider` when CDI/Weld is unavailable, and refactored sequence-related methods to consistently use a new `getSequenceAction()` helper for robust dependency resolution. [[1]](diffhunk://#diff-7207aa8168d5d201e0c9247e22c7d6f3862c6126b158b8155a5aa8877ed20139R110-R127) [[2]](diffhunk://#diff-7207aa8168d5d201e0c9247e22c7d6f3862c6126b158b8155a5aa8877ed20139L351-R451) [[3]](diffhunk://#diff-7207aa8168d5d201e0c9247e22c7d6f3862c6126b158b8155a5aa8877ed20139L409-R476) [[4]](diffhunk://#diff-7207aa8168d5d201e0c9247e22c7d6f3862c6126b158b8155a5aa8877ed20139L432-R501)
* Enhanced `WeldUtils` to handle cases where `ServletContext` or `BeanManager` are unavailable, including safer bean instantiation and improved error handling. [[1]](diffhunk://#diff-102846adc003d0049b507204105ac4caaaf0751de8e2ae650773e5a01e172012L55-R61) [[2]](diffhunk://#diff-102846adc003d0049b507204105ac4caaaf0751de8e2ae650773e5a01e172012L85-R91) [[3]](diffhunk://#diff-102846adc003d0049b507204105ac4caaaf0751de8e2ae650773e5a01e172012R115-R128) [[4]](diffhunk://#diff-102846adc003d0049b507204105ac4caaaf0751de8e2ae650773e5a01e172012R161-R163)

Test isolation and cleanup:

* Added explicit JUnit 4/Vintage cleanup hook in `OBBaseTest` to prevent DAL session/context leaks between tests, and ensured compatibility with both JUnit 4 and 5 lifecycle annotations. [[1]](diffhunk://#diff-264418eaef1ddfad39343275f999d457b0517b588b478d0240a6af5e5661cb31R383) [[2]](diffhunk://#diff-264418eaef1ddfad39343275f999d457b0517b588b478d0240a6af5e5661cb31R402) [[3]](diffhunk://#diff-264418eaef1ddfad39343275f999d457b0517b588b478d0240a6af5e5661cb31R431-R472)
* Improved test isolation in `TestCostingBase` by programmatically disabling the automatic price difference correction preference before each test and ensuring costing rule is validated with a start-of-day timestamp. [[1]](diffhunk://#diff-12e31cefbaedea918d2babbba20a8727dfdf165b0982f9bb852fe1275d0ac79fR55-R66) [[2]](diffhunk://#diff-12e31cefbaedea918d2babbba20a8727dfdf165b0982f9bb852fe1275d0ac79fL145-L152)

Costing test accuracy:

* Updated costing tests in `TestCostingSourceAdjustments` to use correct average price calculations and assertions, ensuring accurate validation of costing logic. [[1]](diffhunk://#diff-b9a42852d9de771985ce6a0789e3ef00b33d70a9ea43a7a4c73f0741df1f8364R1054) [[2]](diffhunk://#diff-b9a42852d9de771985ce6a0789e3ef00b33d70a9ea43a7a4c73f0741df1f8364L1082-R1083) [[3]](diffhunk://#diff-b9a42852d9de771985ce6a0789e3ef00b33d70a9ea43a7a4c73f0741df1f8364L1100-R1109) [[4]](diffhunk://#diff-b9a42852d9de771985ce6a0789e3ef00b33d70a9ea43a7a4c73f0741df1f8364L1118-R1119) [[5]](diffhunk://#diff-b9a42852d9de771985ce6a0789e3ef00b33d70a9ea43a7a4c73f0741df1f8364R1302-R1303)


src-test/src/org/openbravo/test/costing/utils/TestCostingUtils.java
```diff
@@ -2504,6 +2528,7 @@ public static Invoice createInvoiceFromOrder(String orderId, boolean issotrx, Bi
         else {
           invoiceLine = (InvoiceLine) DalUtil.copy(invoice.getInvoiceLineList().get(0), false);
           TestCostingUtils.setGeneralData(invoiceLine);
+          invoiceLine.setInvoiceLineTaxList(new ArrayList<>());
           invoiceLine.setInvoice(invoice);
```
etendobot: **Suggestion**

Verify if manually resetting the tax list to an empty ArrayList is necessary, as DalUtil.copy with the 'false' flag already excludes collections from the copy.

```java
// If DalUtil.copy is used with false (non-recursive), collections like invoiceLineTaxList are already uninitialized. If initialization is needed for specific test logic, ensure tax logic is valid.
```

**Rationale**: Setting an empty ArrayList manually on a DAL object might bypass framework-level tax calculations or cause issues if the collection is expected to be managed by Hibernate. Since DalUtil.copy(..., false) already ignores collections, this manual reset should be verified against the test's requirements for tax lines.

src/org/openbravo/costing/CostAdjustmentProcess.java
```diff
@@ -325,38 +334,61 @@ private void generateTransactionCosts(CostAdjustmentLine costAdjustmentLine) {
...
 
   public static CostingAlgorithmAdjustmentImp doGetAlgorithmAdjustmentImp(String strJavaClass) {
-    CostAdjustmentProcess cap = WeldUtils
-        .getInstanceFromStaticBeanManager(CostAdjustmentProcess.class);
+    CostAdjustmentProcess cap = getProcessInstance();
     return cap.getAlgorithmAdjustmentImp(strJavaClass);
   }
+
+  private static CostAdjustmentProcess getProcessInstance() {
+    try {
```
etendobot: **Suggestion**

Consider moving this 'Weld or OBProvider' pattern to a utility class if it is needed in other costing processes.

```java
private static CostAdjustmentProcess getProcessInstance() {
    try {
      return WeldUtils.getInstanceFromStaticBeanManager(CostAdjustmentProcess.class);
    } catch (Exception e) {
      log.debug("Falling back to OBProvider instance for CostAdjustmentProcess: {}", e.getMessage());
      return OBProvider.getInstance().get(CostAdjustmentProcess.class);
    }
  }
```

**Rationale**: This fallback mechanism is a clever way to handle both production (CDI-enabled) and vintage test (non-CDI) environments. It ensures that the process can still be instantiated even when Weld is not available.

src/org/openbravo/costing/CostingUtils.java
```diff
@@ -909,12 +909,12 @@ public static boolean existsProcessedTransactions(Product product,
...
         .createCriteria(MaterialTransaction.class);
-    criteria.createAlias(MaterialTransaction.PROPERTY_STORAGEBIN, "sb")
-        .add(Restrictions.eq(MaterialTransaction.PROPERTY_PRODUCT, product))
+    criteria.add(Restrictions.eq(MaterialTransaction.PROPERTY_PRODUCT, product))
         .add(Restrictions.eq(MaterialTransaction.PROPERTY_ISPROCESSED, true))
         .add(Restrictions.in(MaterialTransaction.PROPERTY_ORGANIZATION + ".id", orgs));
     if (costDimensions.get(CostDimension.Warehouse) != null) {
-      criteria.add(Restrictions.eq("sb." + Locator.PROPERTY_WAREHOUSE + ".id",
+      criteria.add(Restrictions.eq(
+          MaterialTransaction.PROPERTY_STORAGEBIN + "." + Locator.PROPERTY_WAREHOUSE + ".id",
```
etendobot: **Suggestion**

Compare the warehouse entity directly instead of appending '.id'.

```java
criteria.add(Restrictions.eq(
          MaterialTransaction.PROPERTY_STORAGEBIN + "." + Locator.PROPERTY_WAREHOUSE,
          costDimensions.get(CostDimension.Warehouse)));
```

**Rationale**: While the dot-notation property path is valid in Etendo's DAL for simple navigation, using '.id' can sometimes trigger unnecessary joins or redundant SQL depending on the Hibernate version and configuration. Comparing the association property directly with the business object is generally cleaner and better optimized by the DAL.

src-test/src/org/openbravo/test/base/OBBaseTest.java
```diff
@@ -426,6 +428,48 @@ public void testDone(TestInfo testInfo) {
...
+  /**
+   * JUnit 4/Vintage cleanup hook.
+   * <p>
+   * For JUnit 5, cleanup is done by {@link DalCleanupExtension}. Vintage tests do not execute that
+   * extension, so they require this explicit cleanup to avoid leaking DAL session/context between
+   * tests.
+   */
+  @org.junit.After
+  public void vintageTestDone() {
+    try {
```
etendobot: **Suggestion**

Consider using DalThreadCleaner.getInstance().clean(errorOccurred) if available, or ensuring commitAndClose() is called for active transactions when no rollback is requested.

```java
  public void vintageTestDone() {
    boolean errorOccurred = false;
    // ... logic to detect error ...
    try {
      if (SessionHandler.isSessionHandlerPresent()) {
        if (SessionHandler.getInstance().getDoRollback() || errorOccurred) {
          SessionHandler.getInstance().rollback();
        } else if (SessionHandler.getInstance().getSession().getTransaction().isActive()) {
          SessionHandler.getInstance().commitAndClose();
        } else {
          SessionHandler.getInstance().getSession().close();
        }
      }
    } ...
```

**Rationale**: The manual cleanup logic in vintageTestDone should ideally commit successful transactions if no error occurred, similar to how the legacy TestWatcher or the new DalCleanupExtension works. The current implementation only rolls back or closes.

src/org/openbravo/costing/CostAdjustmentProcess.java
```diff
@@ -325,38 +334,61 @@ private void generateTransactionCosts(CostAdjustmentLine costAdjustmentLine) {
...
-      throw new OBException(OBMessageUtils.messageBD("CostAlgorithmWithoutAdjustment"));
+      implementor = getAlgorithmAdjustmentImpWithoutCDI(strJavaClass);
     }
     return implementor;
   }
 
+  private CostingAlgorithmAdjustmentImp getAlgorithmAdjustmentImpWithoutCDI(String strJavaClass) {
+    // Fallback used in non-CDI execution contexts (for example, some Vintage/JUnit4 tests).
+    final List<Class<? extends CostingAlgorithmAdjustmentImp>> knownImplementors = List.of(
+        AverageCostAdjustment.class, StandardCostAdjustment.class);
```
etendobot: **Suggestion**

Consider making the non-CDI fallback more extensible or document that custom algorithms must be added here for vintage test support.

```java
    // No direct replacement; consider using a configuration file or scanning if list grows.
```

**Rationale**: The hardcoded list of implementors in the non-CDI fallback is brittle. If a custom module adds a new Costing Algorithm, it will fail in non-CDI environments (like some tests) because it's not in this list. While acceptable for fixing core tests, it is not extensible.

src/org/openbravo/costing/CostingRuleProcess.java
```diff
@@ -208,9 +208,9 @@ private boolean existsTransactions(final Set<String> naturalOrgs, final Set<Stri
                   "   and p.stocked = true" +
                   "   and p.organization.id in (:porgs)" +
                   "   and exists (" +
-                  "     select 1 from MaterialMgmtMaterialTransaction" +
-                  "      where product = p" +
-                  "        and organization.id in (:childOrgs)" +
+                  "     select 1 from MaterialMgmtMaterialTransaction as trx" +
+                  "      where trx.product = p" +
```
etendobot: **Suggestion**

The addition of the 'trx' alias is correct and follows standard Etendo HQL patterns seen in other costing-related queries.

```java
     select 1 from MaterialMgmtMaterialTransaction as trx
```

**Rationale**: The use of aliases in HQL subqueries is a best practice to avoid ambiguity, especially when the outer query and subquery might reference similar property names. This change improves code maintainability and clarity.

etendobot (changes requested):

src-test/src/org/openbravo/test/costing/TestCostingBase.java
```diff
@@ -50,6 +52,18 @@ public class TestCostingBase extends WeldBaseTest {
 
   @Before
   public void setInitialConfiguration() {
```
etendobot: **Suggestion**

This addition is correct and improves test stability by ensuring a clean state for costing preferences.

```java
  @Before
  public void setInitialConfiguration() {
    try {
      OBContext.setOBContext(TestCostingConstants.ADMIN_USER_ID,
          TestCostingConstants.QATESTING_ROLE_ID, TestCostingConstants.QATESTING_CLIENT_ID,
          TestCostingConstants.SPAIN_ORGANIZATION_ID);
      OBContext.setAdminMode(true);
      TestCostingUtils.disableAutomaticPriceDifferenceCorrectionPreference();
    } finally {
      OBContext.restorePreviousMode();
    }
```

**Rationale**: Improving test isolation by explicitly disabling preferences that might affect costing results is a good practice for stable test execution.

src/org/openbravo/costing/LCMatchingProcess.java
```diff
@@ -138,9 +138,11 @@ private void doChecks(final LandedCostCost lcCost, final JSONObject message) {
       throw new OBException(OBMessageUtils.messageBD("LCCostNoMatchings"));
     }
 
-    // Execute checks added implementing LandedCostProcessCheck interface.
-    for (LCMatchingProcessCheck checksInstance : LCMatchingProcessChecks) {
-      checksInstance.doCheck(lcCost, message);
+    // Execute checks added implementing LCMatchingProcessCheck interface.
+    if (LCMatchingProcessChecks != null) {
+      for (LCMatchingProcessCheck checksInstance : LCMatchingProcessChecks) {
```
etendobot: **Suggestion**

Add a check for .isUnsatisfied() to properly handle CDI Instance state.

```java
if (LCMatchingProcessChecks != null && !LCMatchingProcessChecks.isUnsatisfied()) {
```

**Rationale**: In Weld/CDI, an injected 'Instance' is rarely null even if no implementations are found. The correct way to check if any implementations exist is using the .isUnsatisfied() method. While the null check prevents a NullPointerException, it might not be sufficient to verify if the collection is usable.

src/org/openbravo/costing/CancelCostAdjustment.java
```diff
@@ -84,6 +85,8 @@ protected JSONObject execute(Map<String, Object> parameters, String data) {
 
   public static JSONObject doCancelCostAdjustment(CostAdjustment costAdjustmentOrig)
       throws OBException, JSONException {
+    // Ensure entity is materialized to avoid copying a Hibernate proxy class.
+    costAdjustmentOrig = (CostAdjustment) Hibernate.unproxy(costAdjustmentOrig);
     CostAdjustment costAdjustmentCancel = (CostAdjustment) DalUtil.copy(costAdjustmentOrig, false);
```
etendobot: **Suggestion**

Ensure that the Hibernate unproxy call is necessary for the specific use case of DalUtil.copy in this transaction.

```java
costAdjustmentOrig = (CostAdjustment) Hibernate.unproxy(costAdjustmentOrig);
```

**Rationale**: Explicitly unproxying a Hibernate entity before using DalUtil.copy() ensures that the underlying implementation class is cloned rather than the proxy wrapper. This is a common pattern in Etendo to avoid ClassCastExceptions or issues when the copied object is used outside its original session context.

src/org/openbravo/costing/LandedCostProcess.java
```diff
@@ -186,8 +186,10 @@ private void doChecks(final LandedCost landedCost, final JSONObject message) {
     }
 
     // Execute checks added implementing LandedCostProcessCheck interface.
-    for (LandedCostProcessCheck checksInstance : landedCostProcessChecks) {
-      checksInstance.doCheck(landedCost, message);
+    if (landedCostProcessChecks != null) {
+      for (LandedCostProcessCheck checksInstance : landedCostProcessChecks) {
```
etendobot: **Suggestion**

Verify if the 'landed_cost_process_checks' variable name in the feedback matches the actual field name 'landedCostProcessChecks'. Ensure that the iteration logic is robust.

```java
if (landedCostProcessChecks != null) {
```

**Rationale**: While it is a good practice to check for nullity, the 'landedCostProcessChecks' field is usually an 'Instance<LandedCostProcessCheck>' injected via Weld (@Inject @Any). In most environments, the Instance object itself is not null, even if it contains no implementations. However, adding the check prevents potential NullPointerExceptions if the injection fails or in specific test mock scenarios.

modules_core/org.openbravo.advpaymentmngt/src/org/openbravo/advpaymentmngt/utility/FIN_Utility.java
```diff
@@ -105,7 +107,24 @@ public class FIN_Utility {
...
+      log4j.debug("Falling back to OBProvider instance for FIN_Utility: {}", e.getMessage());
+      return OBProvider.getInstance().get(FIN_Utility.class);
+    }
+  }
+
+  private static FIN_SequenceActionInterface getSequenceAction() {
+    try {
+      FIN_Utility utility = getInstance();
+      if (utility.sequenceAction != null && !utility.sequenceAction.isUnsatisfied()) {
+        return utility.sequenceAction.get();
```
etendobot: **Suggestion**

Consider using `sequenceAction.isResolvable()` if using a newer Weld version, though `!isUnsatisfied()` is standard for the current Etendo core version.

```java
if (utility.sequenceAction != null && !utility.sequenceAction.isUnsatisfied()) {
```

**Rationale**: The check correctly handles the Weld Instance object to avoid NullPointerExceptions or UnsatisfiedResolutionExceptions when the sequence action is not injected.

modules_core/org.openbravo.base.weld/src/org/openbravo/base/weld/WeldUtils.java
```diff
@@ -82,8 +87,8 @@ public static BeanManager getStaticInstanceBeanManager() {
             staticBeanManager = cdi.getBeanManager();
             log.debug("BeanManager obtained via CDI.current() fallback.");
           } catch (IllegalStateException ise) {
-            log.error("Could not obtain BeanManager via ServletContext, JNDI, or CDI.current(): {}", ise.getMessage());
-            throw new OBException("BeanManager not available, aborting CDI initialization", ise);
+            log.error("Could not obtain BeanManager via ServletContext, JNDI, or CDI.current(): {}",
+                ise.getMessage());
```
etendobot: > [!WARNING]
**⚠️ Blocking Issue**

Restore the 'throw new OBException' to ensure the system fails fast when the CDI container is unavailable.

```java
            log.error("Could not obtain BeanManager via ServletContext, JNDI, or CDI.current(): {}",
                ise.getMessage());
            throw new OBException("BeanManager not available, aborting CDI initialization", ise);
          }
```

**Rationale**: Swallowing the IllegalStateException when the BeanManager cannot be retrieved is a critical failure. If the BeanManager is null, the application will operate in an inconsistent state, leading to unpredictable NullPointerExceptions. Critical framework failures must be propagated using OBException to ensure the system fails fast.

src-test/src/org/openbravo/test/base/OBBaseTest.java
```diff
@@ -380,6 +380,7 @@ public void afterEach(ExtensionContext context) throws Exception {
    * @see TestLogAppender
    */
   @BeforeAll
+  @org.junit.BeforeClass
```
etendobot: **Suggestion**

Ensure that the dual annotation pattern is consistently applied to all initialization hooks if the goal is full Vintage support.

```java
  @BeforeAll
  @org.junit.BeforeClass
  public static void classSetUp() throws Exception {
```

**Rationale**: Adding @BeforeClass ensures compatibility for legacy tests (JUnit 4/Vintage) that extend OBBaseTest but are not yet migrated to JUnit 5 (@BeforeAll).

src/org/openbravo/costing/CancelCostAdjustment.java
etendobot: **Suggestion**

This is a correct fix for proxy-related issues during object cloning. Ensure that Hibernate (org.hibernate.Hibernate) is correctly imported.

```java
costAdjustmentOrig = (CostAdjustment) Hibernate.unproxy(costAdjustmentOrig);
```

**Rationale**: The use of Hibernate.unproxy is a valid way to ensure the actual entity is passed to DalUtil.copy instead of a Hibernate proxy. This prevents potential issues where the proxy's class name (containing '_$$_hibernate_proxy_') causes failures in the copy logic which relies on entity mapping. While not strictly required if DAL utilities like DalUtil.getId() are used, it is a common fix for cases where the full object structure must be cloned.

src/org/openbravo/costing/CostAdjustmentProcess.java
etendobot: **Suggestion**

The fallback mechanism in getProcessInstance() is a robust way to ensure compatibility with non-CDI test environments.

```java
  private static CostAdjustmentProcess getProcessInstance() {
    try {
      return WeldUtils.getInstanceFromStaticBeanManager(CostAdjustmentProcess.class);
    } catch (Exception e) {
      log.debug("Falling back to OBProvider instance for CostAdjustmentProcess: {}", e.getMessage());
      return OBProvider.getInstance().get(CostAdjustmentProcess.class);
    }
  }
```

**Rationale**: WeldUtils.getInstanceFromStaticBeanManager fails if the CDI container is not initialized. Providing a fallback to OBProvider.getInstance().get() allows the code to run in legacy test environments while maintaining CDI support in production.

src/org/openbravo/costing/LandedCostDistributionByAmount.java
```diff
@@ -113,7 +113,7 @@ public void distributeAmount(final LandedCostCost lcCost, final boolean isMatchi
       }
       pendingAmt = pendingAmt.subtract(receiptAmt);
       final LCReceipt lcrl = (LCReceipt) OBDal.getInstance()
-          .getProxy(LCReceipt.ENTITY_NAME, receiptCosts.get());
+          .getProxy(LCReceipt.ENTITY_NAME, row[0]);
       final LCReceiptLineAmt lcrla = OBProvider.getInstance().get(LCReceiptLineAmt.class);
```
etendobot: > [!WARNING]
**⚠️ Blocking Issue**

Declare and initialize the 'row' variable by assigning 'receiptCosts.get()' to it at the beginning of the loop iteration.

```java
final Object[] row = receiptCosts.get();\n      final LCReceipt lcrl = (LCReceipt) OBDal.getInstance()\n          .getProxy(LCReceipt.ENTITY_NAME, row[0]);
```

**Rationale**: The variable 'row' is used but has not been declared or assigned the value of 'receiptCosts.get()'. Using an undefined variable leads to compilation failure. Additionally, calling '.get()' once and storing the result in a variable is the standard practice for ScrollableResults to ensure consistency and performance.

src/org/openbravo/erpCommon/ad_forms/AcctServer.java
```diff
@@ -1468,10 +1468,9 @@ public boolean isConvertible(AcctSchema acctSchema, ConnectionProvider conn)
             OBCriteria<Currency> currencyCrit = OBDal.getInstance().createCriteria(Currency.class);
             currencyCrit.add(Restrictions.eq(Currency.PROPERTY_ID, acctSchema.m_C_Currency_ID));
             currencyCrit.setProjection(Projections.max(Currency.PROPERTY_STANDARDPRECISION));
-            Long precision = 0L;
-            if (currencyCrit.count() > 0) {
-              List<Long> toCurrency = currencyCrit.list(Long.class);
-              precision = toCurrency.get(0);
+            Long precision = currencyCrit.uniqueResult(Long.class);
+            if (precision == null) {
```
etendobot: **Suggestion**

Consider adding .setMaxResults(1) before .uniqueResult() for defensive coding.

```java
Long precision = currencyCrit.setMaxResults(1).uniqueResult(Long.class);
```

**Rationale**: When using uniqueResult() with a criteria that could potentially return multiple results (even if logically unexpected), it is a best practice to set max results to 1 to avoid a NonUniqueResultException. However, since this query filters by a specific ID (Currency.PROPERTY_ID), the risk is minimal.

src/org/openbravo/materialmgmt/InventoryCountProcess.java
```diff
@@ -373,9 +370,9 @@ private List<InventoryCountLine> getLinesWithDuplicatedProducts(final InventoryC
                   "         from MaterialMgmtInventoryCountLine as icl2" +
                   "        where icl.physInventory = icl2.physInventory" +
                   "          and icl.product = icl2.product" +
-                  "          and coalesce(icl.attributeSetValue, '0') = coalesce(icl2.attributeSetValue, '0')" +
-                  "          and coalesce(icl.orderUOM, '0') = coalesce(icl2.orderUOM, '0')" +
-                  "          and coalesce(icl.uOM, '0') = coalesce(icl2.uOM, '0')" +
+                  "          and coalesce(icl.attributeSetValue.id, '0') = coalesce(icl2.attributeSetValue.id, '0')" +
+                  "          and coalesce(icl.orderUOM.id, '0') = coalesce(icl2.orderUOM.id, '0')" +
+                  "          and coalesce(icl.uOM.id, '0') = coalesce(icl2.uOM.id, '0')" +
```
etendobot: > [!WARNING]
**⚠️ Blocking Issue**

Use explicit LEFT JOINs for 'attributeSetValue' to ensure that lines with null attributes are not incorrectly filtered out by the implicit inner join.

```java
          "   left join icl.attributeSetValue as asi1 " +
          "   left join icl2.attributeSetValue as asi2 " +
          "   where ... " +
          "   and coalesce(asi1.id, '0') = coalesce(asi2.id, '0') " +
```

**Rationale**: In HQL, accessing 'icl.attributeSetValue.id' when 'attributeSetValue' is a nullable relationship implicitly creates an inner join. If the attribute is null, that inventory line will be excluded from the query results entirely instead of being treated as '0'. To correctly compare lines that might have null attributes, you must use explicit LEFT JOINs or a join-less ID check if available, though LEFT JOIN is the safest way to handle coalesce on entity properties.

PR #974, opened 2026-04-01
Epic ETP-3504: Add PATCH method support to WebService framework

valenvivaldi:


src-test/src/org/openbravo/service/web/WebServicePatchTest.java
```diff
@@ -0,0 +1,263 @@
...
+  public void testWebServiceWithPatchSupportDoesNotReturn405() throws Exception {
+    WebService patchCapable = new StubWebService() {
+      @Override
+      public void doPatch(String path, HttpServletRequest request, HttpServletResponse response) {
+        // Successful PATCH handling — no 405
+      }
+    };
+
+    patchCapable.doPatch(PATH_SEPARATOR + "test", mockRequest, mockResponse);
+  }
```
etendobot: **Suggestion**

Add a verification to ensure the 405 status is not set.

```java
    patchCapable.doPatch(PATH_SEPARATOR + "test", mockRequest, mockResponse);
    verify(mockResponse, never()).setStatus(HttpServletResponse.SC_METHOD_NOT_ALLOWED);
```

**Rationale**: The test 'testWebServiceWithPatchSupportDoesNotReturn405' currently passes if no exception is thrown, but it doesn't explicitly verify that the side effect of the default implementation (setting 405 status) was avoided. Adding a verification makes the test's intent clearer and more robust.

src/org/openbravo/service/web/WebServiceServlet.java
```diff
@@ -106,6 +106,22 @@ public void doPut(HttpServletRequest request, HttpServletResponse response)
...
 
+  @Override
+  protected void doPatch(HttpServletRequest request, HttpServletResponse response)
+      throws IOException, ServletException {
+    try {
+      final String segment = WebServiceUtil.getInstance().getFirstSegment(request.getPathInfo());
+      final WebService ws = getWebService(request);
+      if (ws != null) {
+        ws.doPatch(getRemainingPath(request.getPathInfo(), segment), request, response);
+      }
```
etendobot: > [!WARNING]
**⚠️ Blocking Issue**

Verify that the `WebService` interface has been updated to include the `doPatch` method. If not, add it to the interface first.

```java
// Ensure the WebService interface is updated first:
// public void doPatch(String path, HttpServletRequest request, HttpServletResponse response) throws Exception;

ws.doPatch(getRemainingPath(request.getPathInfo(), segment), request, response);
```

**Rationale**: The `WebService` interface (org.openbravo.service.web.WebService) does not contain a `doPatch` method. This call will result in a compilation error unless the interface is also updated in this PR. As the interface change is not included in the provided diff, this is a blocking issue.

PR #1014, opened 2026-05-08
Hotfix ETP-3915: Add unit and integration tests for core classes

Emi-Polliotti:


src-test/src/org/openbravo/base/exception/OBSecurityExceptionTest.java
```diff
@@ -0,0 +1,91 @@
...
+  }
+
+  @Test
+  @DisplayName("constructor with message and cause stores both")
+  void testMessageAndCauseConstructor() {
+    Throwable cause = new IllegalStateException("root cause");
+    OBSecurityException ex = new OBSecurityException(ACCESS_DENIED, cause);
+    assertEquals(ACCESS_DENIED, ex.getMessage());
+    assertNotNull(ex.getCause());
+  }
```
etendobot, line 58: **Suggestion**

Use assertSame(cause, ex.getCause()) for better precision.

```java
    assertSame(cause, ex.getCause());
```

**Rationale**: Using assertSame instead of assertNotNull ensures that the exact cause object provided to the constructor is the one being stored and returned. This provides a more precise verification of the constructor's behavior.

src-test/src/org/openbravo/client/application/process/BaseProcessActionHandlerJupiterTest.java
```diff
@@ -0,0 +1,99 @@
...
+
+/**
+ * Additional JUnit Jupiter tests for {@link BaseProcessActionHandler}.
+ * Complements the existing JUnit 4 BaseProcessActionHandlerTest.
+ */
+@DisplayName("BaseProcessActionHandler (Jupiter)")
+public class BaseProcessActionHandlerJupiterTest {
+
+  private static final String REFRESH_PARENT = "refreshParent";
+
```
etendobot, line 41: **Suggestion**

Remove the local REFRESH_PARENT constant and use BaseProcessActionHandler.REFRESH_PARENT throughout the test class.

```java
// Use BaseProcessActionHandler.REFRESH_PARENT directly in the tests
```

**Rationale**: The constant 'refreshParent' is already defined as a public static field in BaseProcessActionHandler. Redefining it in the test class creates duplication and potential for inconsistency if the core constant name ever changes.

src-test/src/org/openbravo/base/BaseCoreTest.java
```diff
@@ -0,0 +1,120 @@
...
+    when(obContext.getCurrentClient()).thenReturn(client);
+    when(obContext.getCurrentOrganization()).thenReturn(organization);
+    when(obContext.getUser()).thenReturn(user);
+    when(obContext.getRole()).thenReturn(role);
+    when(obContext.getOrganizationStructureProvider()).thenReturn(orgStructureProvider);
+  }
+
+  @AfterEach
+  void baseTearDown() {
+    obDalMock.close();
```
etendobot, line 99: **Suggestion**

Add null checks for 'MockedStatic' objects in the 'baseTearDown' method to prevent NullPointerExceptions during cleanup if the setup fails partially.

```java
  @AfterEach
  void baseTearDown() {
    if (obDalMock != null) {
      obDalMock.close();
    }
    if (obContextMock != null) {
      obContextMock.close();
    }
    if (obProviderMock != null) {
      obProviderMock.close();
    }
    if (obPropertiesProviderMock != null) {
      obPropertiesProviderMock.close();
    }
  }
```

**Rationale**: If an exception occurs during the '@BeforeEach' setup before all 'MockedStatic' objects are initialized, the teardown method might throw a NullPointerException when attempting to close a null reference. Adding null checks ensures a clean teardown even in case of partial initialization.

src-test/src/org/openbravo/base/BaseCoreTest.java
```diff
@@ -0,0 +1,120 @@
...
+import org.openbravo.model.ad.access.User;
+import org.openbravo.model.ad.system.Client;
+import org.openbravo.model.common.enterprise.Organization;
+
+/**
+ * Base class for core Etendo class unit tests. Provides common mock setup
+ * for OBDal, OBContext, OBProvider and OBPropertiesProvider singletons.
+ */
+@MockitoSettings(strictness = Strictness.LENIENT)
+public abstract class BaseCoreTest {
```
etendobot, line 50: **Suggestion**

Consider expanding the Javadoc to include a brief example or more details on how to use the provided mocks in subclasses.

```java
/**
 * Base class for core Etendo class unit tests. Provides common mock setup
 * for OBDal, OBContext, OBProvider and OBPropertiesProvider singletons.
 * 
 * Usage:
 * Extend this class and use the provided mocks (obDal, obContext, etc.) 
 * to define specific behavior for your test case.
 */
@MockitoSettings(strictness = Strictness.LENIENT)
public abstract class BaseCoreTest {
```

**Rationale**: Providing clear usage instructions in the Javadoc helps other developers understand how to extend the base class and what mocks are available.

src-test/src/org/openbravo/advpaymentmngt/utility/FINUtilityTest.java
```diff
@@ -0,0 +1,290 @@
...
+  private static final String DATETIME_FORMAT = "dd-MM-yyyy HH:mm:ss";
+  private static final String DECIMAL_FORMAT = "#,##0.00";
+
+  @Nested
+  @DisplayName("getDate")
+  class GetDateTests {
+    @Test
+    void testGetDateValid() {
+      mockDateFormats(DATE_FORMAT, DATETIME_FORMAT);
+      Date result = FIN_Utility.getDate("15-03-2026");
```
etendobot, line 56: > [!WARNING]
**⚠️ Blocking Issue**

Ensure that all helper methods used in the test are either defined within the test class, imported from a utility class, or provided by the base test class. Consider using MockedStatic for core utility classes like OBDateUtils if the goal is to control date parsing behavior.

```java
// Define mockDateFormats or use MockedStatic<OBDateUtils>
try (MockedStatic<OBDateUtils> mockedDateUtils = mockStatic(OBDateUtils.class)) {
    mockedDateUtils.when(() -> OBDateUtils.getDateFormat()).thenReturn(new SimpleDateFormat(DATE_FORMAT));
    mockedDateUtils.when(() -> OBDateUtils.getDateTimeFormat()).thenReturn(new SimpleDateFormat(DATETIME_FORMAT));
    // ...
}
```

**Rationale**: The method 'mockDateFormats' is not defined in this class or inherited from a known standard base class, nor is it imported. This will cause a compilation error. Tests should be self-contained or use established helpers.

src-test/src/org/openbravo/advpaymentmngt/utility/FINUtilityTest.java
```diff
@@ -0,0 +1,290 @@
...
+
+  private static final String DATE_FORMAT = "dd-MM-yyyy";
+  private static final String DATETIME_FORMAT = "dd-MM-yyyy HH:mm:ss";
+  private static final String DECIMAL_FORMAT = "#,##0.00";
+
+  @Nested
+  @DisplayName("getDate")
+  class GetDateTests {
+    @Test
+    void testGetDateValid() {
```
etendobot, line 54: **Suggestion**

Modify the test to call the method with a null or empty format string to verify the fallback mechanism. If the fallback relies on external components like UIDefinitionController, ensure they are mocked appropriately.

```java
    @Test
    void testEmptyFormatFallsBack() {
      // Test the actual fallback by passing null or empty javaFormat
      String result = FIN_Utility.formatNumber(new BigDecimal("100"), null, ".", ",");
      // Assert based on expected fallback behavior (e.g. UIDefinitionController or default format)
    }
```

**Rationale**: The test name and comment suggest it is testing the fallback logic when javaFormat is empty, but the code passes a specific format ('#0.00'), thereby bypassing the logic it claims to test.

src-test/src/org/openbravo/advpaymentmngt/utility/FINUtilityTest.java
```diff
@@ -0,0 +1,290 @@
...
+import org.openbravo.service.db.DbUtility;
+
+/**
+ * Tests for the {@link FIN_Utility} class covering date parsing, map construction,
+ * exception message extraction, date arithmetic, deposit/payment amount calculation,
+ * and number formatting.
+ */
+@DisplayName("FIN_Utility")
+public class FINUtilityTest extends BaseCoreTest {
+
```
etendobot, line 45: **Suggestion**

Consider whether the tests require a full database/Weld environment. If not, converting to a pure unit test using @ExtendWith(MockitoExtension.class) will improve test execution speed.

```java
public class FINUtilityTest {
  // Use @ExtendWith(MockitoExtension.class) if DAL is not needed
```

**Rationale**: Extending BaseCoreTest (or OBBaseTest) initializes the full Data Access Layer (DAL) and Weld environment, which is significantly slower than a standard JUnit/Mockito unit test. Since most tests here mock static dependencies, the DAL may not be strictly necessary for all scenarios.

src-test/src/org/openbravo/advpaymentmngt/dao/AdvPaymentMngtDaoTest.java
```diff
@@ -0,0 +1,109 @@
...
+
+      assertSame(payment, dao.getObject(FIN_Payment.class, "PAY1"));
+      verify(obDal).get(FIN_Payment.class, "PAY1");
+    }
+
+    @Test
+    void testReturnsNullWhenNotFound() {
+      when(obDal.get(FIN_Payment.class, "NONEXISTENT")).thenReturn(null);
+      assertEquals(null, dao.getObject(FIN_Payment.class, "NONEXISTENT"));
+    }
```
etendobot, line 107: **Suggestion**

Replace assertEquals(null, ...) with assertNull(...) for checking null values.

```java
      assertNull(dao.getObject(FIN_Payment.class, "NONEXISTENT"));
```

**Rationale**: Using assertNull() is more idiomatic and provides clearer intent than assertEquals(null, ...). It also produces better error messages in case of failure.

src-test/src/org/openbravo/erpCommon/utility/DimensionDisplayUtilityIntegrationTest.java
```diff
@@ -0,0 +1,105 @@
...
+    OBContext.setAdminMode(true);
+    try {
+      Client fbClient = OBDal.getInstance().get(Client.class, TestConstants.Clients.FB_GRP);
+      Map<String, String> config = DimensionDisplayUtility
+          .getAccountingDimensionConfiguration(fbClient);
+
+      // Should contain at least one dimension type
+      boolean hasDimension = config.keySet().stream()
+          .anyMatch(k -> k.contains("$Element"));
+      assertTrue("Configuration should contain dimension element keys", hasDimension);
```
etendobot, line 67: **Suggestion**

Use the constant 'DimensionDisplayUtility.ELEMENT' instead of the hardcoded string '$Element'.

```java
.anyMatch(k -> k.contains(DimensionDisplayUtility.ELEMENT));
```

**Rationale**: Using the constant defined in the utility class instead of a hardcoded string makes the code more maintainable and less prone to errors if the prefix changes.

src-test/src/org/openbravo/erpCommon/utility/DimensionDisplayUtilityIntegrationTest.java
```diff
@@ -0,0 +1,105 @@
...
+ */
+public class DimensionDisplayUtilityIntegrationTest extends OBBaseTest {
+
+  @Test
+  public void testGetAccountingDimensionConfigurationReturnsNonNull() {
+    setSystemAdministratorContext();
+    OBContext.setAdminMode(true);
+    try {
+      Client fbClient = OBDal.getInstance().get(Client.class, TestConstants.Clients.FB_GRP);
+      assertNotNull("F&B client should exist", fbClient);
```
etendobot, line 44: **Suggestion**

Add an explicit null check for 'fbClient' in all test methods, consistent with 'testGetAccountingDimensionConfigurationReturnsNonNull'.

```java
Client fbClient = OBDal.getInstance().get(Client.class, TestConstants.Clients.FB_GRP);\n      assertNotNull("F&B client should exist", fbClient);
```

**Rationale**: While the test will fail if the client is null, explicit assertions provide clearer diagnostic messages in test reports when the prerequisite data is missing.

src-test/src/org/openbravo/erpCommon/utility/OBCurrencyUtilsIntegrationTest.java
```diff
@@ -0,0 +1,105 @@
...
+  // F&B US West Coast — no direct currency, inherits from legal entity US
+  private static final String ORG_US_WEST = TestConstants.Orgs.US_WEST;
+
+  @Test
+  public void testOrgWithDirectCurrencyReturnsIt() {
+    setSystemAdministratorContext();
+    String currency = OBCurrencyUtils.getOrgCurrency(ORG_US);
+    assertNotNull("US org should have a currency", currency);
+    assertEquals("100", currency); // USD
+  }
```
etendobot, line 51: **Suggestion**

Consider using constants for currency IDs instead of hardcoded strings.

```java
// If a constant exists in TestConstants or similar:\nassertEquals(TestConstants.Currencies.USD, currency);
```

**Rationale**: While \"100\" and \"102\" are standard IDs in the F&B sample data, using constants improves readability and maintainability of the tests.

src-test/src/org/openbravo/client/kernel/KernelUtilsIntegrationTest.java
```diff
@@ -0,0 +1,135 @@
...
+import org.openbravo.test.base.OBBaseTest;
+
+/**
+ * Integration tests for {@link KernelUtils}.
+ * Requires a running database with the Etendo model loaded.
+ */
+public class KernelUtilsIntegrationTest extends OBBaseTest {
+
+  @Test
+  public void testGetInstanceReturnsNonNull() {
```
etendobot, line 44: **Suggestion**

Centralize the context setup in a '@Before' method.

```java
  @Override
  @Before
  public void setUp() throws Exception {
    super.setUp();
    setSystemAdministratorContext();
  }

  @Test
  public void testGetInstanceReturnsNonNull() {
    assertNotNull(KernelUtils.getInstance());
  }
```

**Rationale**: Repeating 'setSystemAdministratorContext()' in each test method is redundant. Moving it to an overridden 'setUp' method (calling super.setUp()) follows the DRY principle and ensures the context is consistently set before every test execution.

src-test/src/org/openbravo/client/kernel/KernelUtilsIntegrationTest.java
```diff
@@ -0,0 +1,135 @@
...
+    assertNotNull(modules);
+    assertFalse("Module list should not be empty", modules.isEmpty());
+  }
+
+  @Test
+  public void testCoreModuleIsFirst() {
+    setSystemAdministratorContext();
+    List<Module> modules = KernelUtils.getInstance().getModulesOrderedByDependency();
+    assertEquals("Core module should be first", "0", modules.get(0).getId());
+  }
```
etendobot, line 62: **Suggestion**

Perform an emptiness check on the list before accessing the first element.

```java
    assertFalse("The list of modules should not be empty", modules.isEmpty());
    assertEquals("Core module should be first", "0", modules.get(0).getId());
```

**Rationale**: While highly unlikely in an integration test, accessing index 0 directly without a size check can cause an 'IndexOutOfBoundsException'. Using 'assertFalse(modules.isEmpty())' or 'assertTrue(modules.size() > 0)' provides a clearer failure message if the assumption fails.

src-test/src/org/openbravo/erpCommon/ad_callouts/SimpleCalloutCalloutInfoTest.java
```diff
@@ -0,0 +1,478 @@
...
+      SimpleCallout failingCallout = new SimpleCallout() {
+        @Override
+        protected void execute(CalloutInfo ci) throws ServletException {
+          ci.addResult("inpbefore", "ok");
+          throw new ServletException("Callout failed");
+        }
+      };
+
+      try {
+        info.executeCallout(failingCallout);
```
etendobot, line 468: **Suggestion**

Use assertThrows instead of a try-catch block for asserting expected exceptions.

```java
      assertThrows(ServletException.class, () -> info.executeCallout(failingCallout));
```

**Rationale**: Using assertThrows is the standard and more readable way to assert exceptions in JUnit 5.

src-test/src/org/openbravo/erpCommon/ad_callouts/SimpleCalloutCalloutInfoTest.java
```diff
@@ -0,0 +1,478 @@
...
+  }
+
+  // ── Context accessors ────────────────────────────────────────────────
+
+  @Nested
+  @DisplayName("Context accessors")
+  class ContextAccessors {
+
+    @Test
+    void testGetLastFieldChanged() {
```
etendobot, line 85: **Suggestion**

Use a BigDecimal instance instead of a casted String to test the addResult(String, Object) method.

```java
    @Test
    void testAddObjectResult() throws Exception {
      BigDecimal amount = new BigDecimal("100.50");
      info.addResult(INP_AMOUNT, amount);
      JSONObject result = info.getJSONObjectResult();
      assertEquals(amount.toString(), result.getJSONObject(INP_AMOUNT).get(CalloutConstants.VALUE));
    }
```

**Rationale**: The addResult(String, Object) method is primarily intended for types like BigDecimal. Using a BigDecimal in the test better reflects the intended use case than a casted String.

src-test/src/org/openbravo/erpCommon/utility/SequenceIdDataTest.java
```diff
@@ -0,0 +1,74 @@
...
+    String uuid = SequenceIdData.getUUID();
+    assertEquals(uuid, uuid.toUpperCase());
+  }
+
+  @Test
+  @DisplayName("contains no hyphens")
+  void testNoHyphens() {
+    assertNotNull(SequenceIdData.getUUID());
+    assertTrue(!SequenceIdData.getUUID().contains("-"));
+  }
```
etendobot, line 51: **Suggestion**

Use a local variable to store the generated UUID and use `assertFalse` instead of `assertTrue(!...)`.

```java
String uuid = SequenceIdData.getUUID();
    assertFalse(uuid.contains("-"), "UUID should not contain hyphens");
```

**Rationale**: Calling `getUUID()` twice in the same test method results in testing two different instances. While unlikely to fail for this specific case, storing the value in a variable is a more robust testing pattern. Also, `assertFalse` is more idiomatic and readable than `assertTrue(!...)`.

src-test/src/org/openbravo/erpCommon/utility/OBCurrencyUtilsTest.java
```diff
@@ -0,0 +1,121 @@
...
+import org.openbravo.model.common.enterprise.Organization;
+
+/**
+ * Unit tests for {@link OBCurrencyUtils}.
+ */
+@DisplayName("OBCurrencyUtils")
+public class OBCurrencyUtilsTest extends BaseCoreTest {
+
+  @Test
+  void testNullOrgId() {
```
etendobot, line 41: **Suggestion**

Group the null, empty, and blank string checks using JUnit 5 ParameterizedTest.

```java
  @ParameterizedTest\n  @NullAndEmptySource\n  @ValueSource(strings = {\"   \"})\n  void testBlankOrNullOrgId(String orgId) {\n    assertNull(OBCurrencyUtils.getOrgCurrency(orgId));\n  }
```

**Rationale**: Using Parameterized Tests for boundary cases like null, empty, and blank strings reduces boilerplate and improves test clarity.

etendobot, line 41: **Suggestion**

Add a test case to cover the logic when orgId is \"0\".

```java
  @Test\n  void testOrgZeroFallback() {\n    Organization org = mock(Organization.class);\n    when(org.getClient()).thenReturn(client);\n    when(client.getId()).thenReturn(\"CLIENT0\");\n    when(obDal.get(Organization.class, \"0\")).thenReturn(org);\n\n    try (MockedStatic<Utility> utilityMock = mockStatic(Utility.class)) {\n      utilityMock.when(() -> Utility.stringBaseCurrencyId(any(), anyString()))\n          .thenReturn(\"EUR\");\n      assertEquals(\"EUR\", OBCurrencyUtils.getOrgCurrency(\"0\"));\n    }\n  }
```

**Rationale**: The implementation of OBCurrencyUtils.getOrgCurrency contains logic specific to orgId \"0\" (system organization). Testing this case ensures that the branch using the current client ID is correctly exercised.

src-test/src/org/openbravo/test/CoverageTestSuite.java
```diff
@@ -98,7 +98,18 @@
 import org.openbravo.erpCommon.ad_actionButton.CreateWorkEffortTest;
 import org.openbravo.erpCommon.ad_actionButton.EditCCPMeasureValuesTest;
 import org.openbravo.erpCommon.ad_actionButton.ExportReferenceDataTest;
+import org.openbravo.advpaymentmngt.dao.AdvPaymentMngtDaoTest;
+import org.openbravo.advpaymentmngt.utility.FINUtilityTest;
```
etendobot, line 102: **Suggestion**

Reorganize the new imports to follow alphabetical order by package and class name. For example, 'org.openbravo.advpaymentmngt' should be at the top of the 'org.openbravo' block, and 'org.openbravo.erpCommon.ad_callouts.SimpleCalloutCalloutInfoTest' should be placed after 'BackgroundProcessClusterConfigTest'.

```java
// Move these imports to their respective alphabetical positions within the existing import blocks.
```

**Rationale**: Maintaining alphabetical order in imports is a standard practice in the Etendo/Openbravo codebase, especially in large files like CoverageTestSuite, to avoid duplication and improve maintainability.

src-test/src/org/openbravo/test/CoverageTestSuite.java
```diff
@@ -230,12 +241,20 @@
     // advpaymentmngt.utility
     APRMSQLFunctionRegisterTest.class,
     APRM_MatchingUtilityTest.class,
+    FINUtilityTest.class,
+
+    // advpaymentmngt.dao — core coverage
+    AdvPaymentMngtDaoTest.class,
```
etendobot, line 251: **Suggestion**

Remove the '— core coverage' suffix from the comment headers and merge the new tests into existing package-based sections.

```java
// Add to the existing advpaymentmngt.dao section if it exists, or use a standard header.
    AdvPaymentMngtDaoTest.class,
```

**Rationale**: The CoverageTestSuite file follows a strict categorization pattern. Adding '— core coverage' suffix to headers is inconsistent with the rest of the file and redundant, as the entire suite is meant for core coverage.

src-test/src/org/openbravo/test/CoverageTestSuite.java
```diff
@@ -292,6 +312,9 @@
     CalloutHttpServletResponseTest.class,
     CalloutServletConfigTest.class,
 
+    // client.kernel — core coverage
+    RequestContextTest.class,
```
etendobot, line 320: **Suggestion**

Merge RequestContextTest into the existing 'client.kernel' section and ensure alphabetical order.

```java
    // client.kernel
    BaseActionHandlerTest.class,
    RequestContextTest.class,
```

**Rationale**: Test classes within each section should be listed alphabetically. RequestContextTest should follow BaseActionHandlerTest.

src-test/src/org/openbravo/test/CoverageTestSuite.java
```diff
@@ -278,6 +297,7 @@
 
     // client.application.process
     BaseProcessActionHandlerTest.class,
+    BaseProcessActionHandlerJupiterTest.class,
```
etendobot, line 305: **Suggestion**

Verify if BaseProcessActionHandlerJupiterTest is a JUnit 5 test. If so, ensure that CoverageTestSuite is updated to use the JUnit Platform runner or another mechanism that supports both versions, otherwise this test will not be executed.

```java
// Ensure the suite runner is compatible with JUnit 5 (Jupiter) or move this test to a JUnit 5 compatible suite.
```

**Rationale**: The CoverageTestSuite uses the JUnit 4 Suite runner (@RunWith(Suite.class)). JUnit 5 (Jupiter) tests are not natively recognized by the JUnit 4 runner and will likely be skipped, failing the goal of adding them to the coverage metrics.

src-test/src/org/openbravo/test/StandaloneTestSuite.java
```diff
@@ -38,6 +38,9 @@
 import org.openbravo.client.application.test.event.DatasourceEventObserver;
 import org.openbravo.client.application.test.event.PersistanceObserver;
 import org.openbravo.client.kernel.CSSMinifier;
+import org.openbravo.client.kernel.KernelUtilsIntegrationTest;
+import org.openbravo.erpCommon.utility.DimensionDisplayUtilityIntegrationTest;
+import org.openbravo.erpCommon.utility.OBCurrencyUtilsIntegrationTest;
```
etendobot, line 45: **Suggestion**

Sort the newly added imports alphabetically with respect to existing ones.

```java
// Move to correct alphabetical position after org.openbravo.email imports
```

**Rationale**: Alphabetical sorting of imports is a standard practice that improves code maintainability.

src-test/src/org/openbravo/test/StandaloneTestSuite.java
```diff
@@ -476,6 +479,11 @@
     // RptMA_ProcessPlan
     RptMA_ProcessPlanTest.class,
     
+    // Core coverage — integration tests
+    OBCurrencyUtilsIntegrationTest.class,
+    DimensionDisplayUtilityIntegrationTest.class,
```
etendobot, line 486: **Suggestion**

Sort the newly added test classes alphabetically within the group.

```java
    DimensionDisplayUtilityIntegrationTest.class,
    KernelUtilsIntegrationTest.class,
    OBCurrencyUtilsIntegrationTest.class,
```

**Rationale**: The StandaloneTestSuite class list generally follows alphabetical order within functional groups. Sorting the new additions maintains this pattern.

etendobot (changes requested):

etendobot (approved):

src-test/src/org/openbravo/client/application/window/ApplicationDictionaryCachedStructuresTest.java
```diff
@@ -1,127 +1,165 @@
...
+import org.openbravo.dal.service.OBDal;
+
+/**
+ * Unit tests for {@link ApplicationDictionaryCachedStructures}.
+ */
+@DisplayName("ApplicationDictionaryCachedStructures Tests")
+class ApplicationDictionaryCachedStructuresTest {
+
+  private static final String IN_DEVELOPMENT_MODULES_FIELD = "inDevelopmentModules";
+  private static final String USE_CACHE_FIELD = "useCache";
```
etendobot: **Suggestion**

Remove the unused constant IN_DEVELOPMENT_MODULES_FIELD.

```java

```

**Rationale**: The constant is defined but not used in the new implementation. The previous version used it for reflection, but the new version uses direct calls to the init() method.

src-test/src/org/openbravo/client/application/window/ApplicationDictionaryCachedStructuresTest.java
```diff
@@ -1,127 +1,165 @@
...
+  private static final String MODULE_ID_UNKNOWN = "UNKNOWN_MODULE_999";
```
etendobot: **Suggestion**

Remove the unused constant USE_CACHE_FIELD.

```java

```

**Rationale**: The constant is defined but not used in the new implementation.

src-test/src/org/openbravo/client/application/window/ApplicationDictionaryCachedStructuresTest.java
```diff
@@ -1,127 +1,165 @@
...
-  @Test
-  public void testUseCacheReturnsFalse() throws Exception {
-    setPrivateField(instance, "useCache", false);
-    assertFalse(instance.useCache());
-  }
   /**
-   * Is in development with dev modules.
-   * @throws Exception if an error occurs
+   * Creates an instance bypassing the constructor, then sets internal fields via reflection.
    */
```
etendobot, line 67: **Suggestion**

Update the Javadoc to accurately describe the method's behavior.

```java
   * Creates an instance and initializes it with the provided development modules.
```

**Rationale**: The Javadoc description does not match the actual implementation of the method, which uses standard instantiation and mocks the DAL layer instead of using reflection.

src-test/src/org/openbravo/client/application/window/ApplicationDictionaryCachedStructuresTest.java
```diff
@@ -1,127 +1,165 @@
...
+  private ApplicationDictionaryCachedStructures createInstanceWithModules(
+      Set<String> devModules) throws Exception {
```
etendobot: **Suggestion**

Remove the unnecessary 'throws Exception' from the method signature.

```java
  private ApplicationDictionaryCachedStructures createInstanceWithModules(
      Set<String> devModules) {
```

**Rationale**: The method implementation does not throw any checked exceptions, so the 'throws Exception' clause is unnecessary.

src-test/src/org/openbravo/client/application/report/ReportingUtilsTest.java
```diff
@@ -0,0 +1,377 @@
...
+  // correctMaskForGrouping (private method, tested via reflection)
+  // ---------------------------------------------------------------
+  @Nested
+  @DisplayName("correctMaskForGrouping() via reflection")
+  class CorrectMaskForGroupingTests {
+
+    private String invokeCorrectMaskForGrouping(String mask, String decimalSymbol,
+        String groupingSymbol) throws Exception {
+      Method method = ReportingUtils.class.getDeclaredMethod(
+          "correctMaskForGrouping", String.class, String.class, String.class);
```
etendobot: **Suggestion**

Consider making 'correctMaskForGrouping' package-private (default visibility) in ReportingUtils.java and annotating it with '@VisibleForTesting' to avoid using reflection in the test class.

```java
// No change required to the test, but consider making the target method package-private and using @VisibleForTesting in ReportingUtils.java
```

**Rationale**: Testing private methods via reflection makes tests dependent on internal implementation details and can be brittle. Package-private visibility is often preferred for methods that need unit testing.

src-test/src/org/openbravo/advpaymentmngt/utility/FINUtilityTest.java
```diff
@@ -0,0 +1,271 @@
...
+  private static final String DATE_FORMAT = "dd-MM-yyyy";
+  private static final String DATETIME_FORMAT = "dd-MM-yyyy HH:mm:ss";
+  private static final String DECIMAL_FORMAT = "#,##0.00";
+
+  // --- getDate tests ---
+
+  @Test
+  public void testGetDateValid() {
+    mockDateFormats(DATE_FORMAT, DATETIME_FORMAT);
+    Date result = FIN_Utility.getDate("15-03-2026");
```
etendobot: > [!WARNING]
**⚠️ Blocking Issue**

Implement the 'mockDateFormats' helper method or replace its calls with the proper MockedStatic configuration for OBDateUtils or similar date utility classes used by FIN_Utility.

```java
try (MockedStatic<OBDateUtils> mockedDateUtils = mockStatic(OBDateUtils.class)) {
  mockedDateUtils.when(() -> OBDateUtils.getDateFormat()).thenReturn(new SimpleDateFormat(DATE_FORMAT));
  mockedDateUtils.when(() -> OBDateUtils.getDateTimeFormat()).thenReturn(new SimpleDateFormat(DATETIME_FORMAT));
  // ... test logic ...
}
```

**Rationale**: The method 'mockDateFormats' is not defined in this class, inherited from 'BaseCoreTest', or imported from any utility class. Calling an undefined method will cause a compilation failure. The developer should either implement the helper method or use MockedStatic to configure date formatting behavior as suggested in the previous review.

src-test/src/org/openbravo/advpaymentmngt/dao/AdvPaymentMngtDaoTest.java
```diff
@@ -0,0 +1,102 @@
...
+
+    assertSame(payment, dao.getObject(FIN_Payment.class, "PAY1"));
+    verify(obDal).get(FIN_Payment.class, "PAY1");
+  }
+
+  @Test
+  public void testReturnsNullWhenNotFound() {
+    when(obDal.get(FIN_Payment.class, "NONEXISTENT")).thenReturn(null);
+    assertEquals(null, dao.getObject(FIN_Payment.class, "NONEXISTENT"));
+  }
```
etendobot, line 108: **Suggestion**

Use 'assertNull' instead of 'assertEquals(null, ...)' for clearer and more idiomatic assertions. Note that this requires adding 'import static org.junit.Assert.assertNull;'.

```java
assertNull(dao.getObject(FIN_Payment.class, "NONEXISTENT"));
```

**Rationale**: Using specific assertions like 'assertNull' improves code readability and provides more descriptive error messages.

src-test/src/org/openbravo/erpCommon/businessUtility/InitialClientSetupTest.java
```diff
@@ -0,0 +1,491 @@
...
+  @DisplayName("getLog Tests")
+  class GetLogTests {
+
+    @Test
+    @DisplayName("getLog returns accumulated header and log content")
+    void getLogReturnsAccumulatedContent() throws Exception {
+      InitialClientSetup setup = new InitialClientSetup();
+
+      Field headerField = InitialClientSetup.class.getDeclaredField("strHeaderLog");
+      headerField.setAccessible(true);
```
etendobot: **Suggestion**

Avoid using reflection to access private fields in tests when possible.

```java
// Consider making these fields package-private and using @VisibleForTesting in the source class to avoid reflection.
// Field headerField = setup.getClass().getDeclaredField("strHeaderLog"); // (Example of existing reflection)
```

**Rationale**: Tests that rely on reflection are fragile because they depend on private implementation details (like field names) that are not part of the stable API. Renaming a field will break the test even if the functionality remains unchanged.

src-test/src/org/openbravo/erpCommon/businessUtility/InitialClientSetupTest.java
```diff
@@ -0,0 +1,491 @@
...
+
+      org.openbravo.model.ad.access.Role mockRole = mock(
+          org.openbravo.model.ad.access.Role.class);
+      Field roleField = InitialClientSetup.class.getDeclaredField("role");
+      roleField.setAccessible(true);
+      roleField.set(setup, mockRole);
+
+      try (MockedStatic<InitialSetupUtility> isuMock = mockStatic(InitialSetupUtility.class)) {
+        isuMock.when(() -> InitialSetupUtility.getLanguage(anyString())).thenReturn(null);
+        isuMock.when(() -> InitialSetupUtility.insertUser(
```
etendobot: **Suggestion**

Use specific argument matchers or values instead of `anyString()` when the expected input is known.

```java
isuMock.when(() -> InitialSetupUtility.getLanguage("en_US")).thenReturn(mockLanguage);
```

**Rationale**: Specific matchers provide better verification that the code under test is passing the correct data to its dependencies. This helps detect bugs where the wrong variable or a hardcoded incorrect value is passed.

src-test/src/org/openbravo/erpCommon/utility/ComboTableDataTest.java
```diff
@@ -0,0 +1,298 @@
...
+    void shouldAddFieldToSelectList() throws Exception {
+      comboTableData.addSelectField("table.column", "alias1");
+
+      Field selectField = ComboTableData.class.getDeclaredField("select");
+      selectField.setAccessible(true);
+      @SuppressWarnings("unchecked")
+      List<Object> selectList = (List<Object>) selectField.get(comboTableData);
+
+      assertNotNull(selectList);
+      assertEquals(1, selectList.size());
```
etendobot: **Suggestion**

Consider verifying the contents of the 'selectList' to ensure the field was added correctly.

```java
      assertNotNull(selectList);
      assertEquals(1, selectList.size());
      // Assuming selectList contains the formatted field
      assertTrue(selectList.get(0).toString().contains("table.column"));
```

**Rationale**: Verifying only the size of the collection does not guarantee that the correct data was processed and stored. Adding content verification makes the test more robust.

src-test/src/org/openbravo/erpCommon/utility/ComboTableDataTest.java
```diff
@@ -0,0 +1,298 @@
...
+  void setUp() {
+    comboTableData = new ComboTableData();
+  }
+
+  @Nested
+  @DisplayName("Constructor tests")
+  class ConstructorTests {
+
+    @Test
+    @DisplayName("Default no-arg constructor should create a non-null instance")
```
etendobot: **Suggestion**

Verify that the parameter was actually added to the internal parameters list in ComboTableData.

```java
    @Test
    @DisplayName("Should replace @SomeParam@ with ? and add parameter")
    void shouldReplaceCustomParamWithQuestionMark() throws Exception {
      String input = "column = @SomeParam@";
      String result = comboTableData.parseContext(input, "WHERE");
      assertTrue(result.contains("?"));
      assertFalse(result.contains("@SomeParam@"));

      Field paramsField = ComboTableData.class.getDeclaredField("parameters");
      paramsField.setAccessible(true);
      List<?> parameters = (List<?>) paramsField.get(comboTableData);
      assertNotNull(parameters);
      assertFalse(parameters.isEmpty(), "Parameter should have been added to the internal list");
    }
```

**Rationale**: The parseContext method is responsible for both replacing placeholders and registering the extracted parameters. Verifying only the string result misses the side effect of parameter registration.

src-test/src/org/openbravo/erpCommon/utility/SequenceIdDataTest.java
```diff
@@ -0,0 +1,66 @@
...
+  public void testUppercase() {
+    String uuid = SequenceIdData.getUUID();
+    assertEquals(uuid, uuid.toUpperCase());
+  }
+
+  @Test
+  public void testNoHyphens() {
+    assertNotNull(SequenceIdData.getUUID());
+    assertTrue(!SequenceIdData.getUUID().contains("-"));
+  }
```
etendobot, line 51: **Suggestion**

Store the UUID in a variable to test the same instance and use 'assertFalse' for better readability.

```java
    String uuid = SequenceIdData.getUUID();
    assertNotNull(uuid);
    assertFalse("UUID should not contain hyphens", uuid.contains("-"));
```

**Rationale**: In the current implementation, 'testNoHyphens' calls 'SequenceIdData.getUUID()' twice, meaning it tests two different UUID instances. It is a better practice to perform all assertions on the same instance. Additionally, 'assertFalse(condition)' is more readable and idiomatic than 'assertTrue(!condition)'.

src-test/src/org/openbravo/erpCommon/businessUtility/InitialOrgSetupTest.java
```diff
@@ -0,0 +1,218 @@
...
+  // ── getOrgId ─────────────────────────────────────────────────────────
+
+  @Nested
+  @DisplayName("getOrgId")
+  class GetOrgIdTests {
+
+    @Test
+    @DisplayName("should return empty string when org is null")
+    void returnsEmptyWhenOrgIsNull() {
+      assertEquals("", setup.getOrgId());
```
etendobot, line 74: **Suggestion**

Consider removing this redundant test case as the functionality is already verified at line 98.

```java
    // Covered in ConstructorTests.orgIdIsEmptyWhenOrgIsNull
```

**Rationale**: The test case for `getOrgId()` when the organization is null is redundant because it is already covered in the `ConstructorTests` nested class (line 98). Removing duplicate tests improves maintenance.

src-test/src/org/openbravo/erpCommon/businessUtility/InitialOrgSetupTest.java
```diff
@@ -0,0 +1,218 @@
...
+
+    @Test
+    @DisplayName("should return error when no CoA modules found")
+    void errorWhenNoCoaModules() {
+      try (MockedStatic<InitialSetupUtility> isuMock = mockStatic(InitialSetupUtility.class)) {
+        isuMock.when(() -> InitialSetupUtility.getCOAModules(anyString()))
+            .thenReturn(Collections.emptyList());
+
+        OBError result = setup.coaModule("");
+        assertEquals("Error", result.getType());
```
etendobot, line 131: **Suggestion**

Verify if the `coaModule` implementation requires additional stubs for static methods like `InitialSetupUtility.getTree` and `OBDal.getInstance().get` to ensure the test correctly reaches the success logic.

```java
        isuMock.when(() -> InitialSetupUtility.getTree(any(), any(), any())).thenReturn(mock(Tree.class));\n        OBError result = setup.coaModule("");
```

**Rationale**: In many versions of the `InitialOrgSetup` class, the `coaModule` method calls `InitialSetupUtility.getTree()` and `OBDal.getInstance().get()` before processing the module list. If these static methods are not stubbed in the test, they may return null or throw an exception, causing the test to fail or return an 'Error' type instead of the expected 'Success'.

src-test/src/org/openbravo/service/db/DataImportServiceTest.java
```diff
@@ -0,0 +1,148 @@
...
+  class ImportDataFromXMLTests {
+
+    @Test
+    @DisplayName("Should throw exception when called with null client, null org and empty XML")
+    void shouldThrowExceptionWithNullClientNullOrgAndEmptyXml() {
+      DataImportService service = new DataImportService();
+      DataImportService.setInstance(service);
+
+      assertThrows(Exception.class, () -> service.importDataFromXML(null, null, ""));
+    }
```
etendobot: **Suggestion**

Use a more specific exception class, such as OBException.class, in the assertThrows assertions.

```java
      assertThrows(OBException.class, () -> service.importDataFromXML(null, null, ""));
```

**Rationale**: Asserting against the generic Exception.class is considered a weak testing practice because it can mask unexpected runtime errors, such as NullPointerException. Since the DataImportService implementation is known to wrap parsing exceptions into OBException, asserting the more specific type ensures the test fails if a different, unexpected exception is thrown.

src-test/src/org/openbravo/service/db/DataImportServiceTest.java
```diff
@@ -0,0 +1,148 @@
...
+  @Nested
+  @DisplayName("importDataFromXML tests")
+  class ImportDataFromXMLTests {
+
+    @Test
+    @DisplayName("Should throw exception when called with null client, null org and empty XML")
+    void shouldThrowExceptionWithNullClientNullOrgAndEmptyXml() {
+      DataImportService service = new DataImportService();
+      DataImportService.setInstance(service);
+
```
etendobot, line 78: **Suggestion**

Remove the redundant calls to DataImportService.setInstance(service) within the ImportDataFromXMLTests nested class.

```java
      // Remove this line
```

**Rationale**: In the context of these specific tests, the 'service' object is being used directly to call the method under test. Setting it as the global singleton instance via setInstance is unnecessary and adds noise to the test setup.

src-test/src/org/openbravo/erpCommon/utility/OBCurrencyUtilsTest.java
```diff
@@ -0,0 +1,119 @@
...
+
+  @Test
+  public void testNullOrgId() {
+    assertNull(OBCurrencyUtils.getOrgCurrency(null));
+  }
+
+  @Test
+  public void testBlankOrgId() {
+    assertNull(OBCurrencyUtils.getOrgCurrency("   "));
+  }
```
etendobot, line 48: **Suggestion**

Consider using an explicit null check before dereferencing the result of `getOrgCurrency`.

```java
Currency result = OBCurrencyUtils.getOrgCurrency("ORG1");
    assertNotNull("The currency should not be null", result);
    assertEquals("The currency ID should match", "102", result.getId());
```

**Rationale**: In JUnit tests, providing an explicit `assertNotNull` before accessing properties of the result helps in generating clearer failure messages compared to a standard `NullPointerException`. This is a best practice for unit testing in the Etendo/Openbravo environment.

src-test/src/org/openbravo/test/CoverageTestSuite.java
```diff
@@ -278,6 +303,10 @@
...
+
+    // client.application.report — core coverage
+    ReportingUtilsTest.class,
```
etendobot: **Suggestion**

Replace the em-dash (—) with a standard hyphen (-) in the comments.

```java
    // client.application.report - core coverage
```

**Rationale**: Using em-dashes (—) or other non-ASCII characters in comments is generally discouraged as it may cause encoding issues in some development environments or terminal outputs. Standard hyphens (-) or double hyphens (--) are preferred.

src-test/src/org/openbravo/test/CoverageTestSuite.java
```diff
@@ -145,6 +157,8 @@
 import org.openbravo.erpCommon.ad_reports.AgingDaoTest;
 import org.openbravo.erpCommon.ad_reports.AgingDataTest;
 import org.openbravo.erpCommon.businessUtility.AccountTreeTest;
+import org.openbravo.erpCommon.businessUtility.InitialClientSetupTest;
+import org.openbravo.erpCommon.businessUtility.InitialOrgSetupTest;
```
etendobot, line 160: **Suggestion**

Organize imports alphabetically. InitialClientSetupTest and InitialOrgSetupTest should come after AuditTrailDeletedRecordsTest.

```java
 import org.openbravo.erpCommon.businessUtility.AuditTrailDeletedRecordsTest;
+import org.openbravo.erpCommon.businessUtility.InitialClientSetupTest;
+import org.openbravo.erpCommon.businessUtility.InitialOrgSetupTest;
```

**Rationale**: Maintaining imports in alphabetical order makes the code easier to navigate and reduces merge conflicts in large test suites like CoverageTestSuite. Initial... comes after Audit... alphabetically.

src-test/src/org/openbravo/test/CoverageTestSuite.java
```diff
@@ -470,9 +516,15 @@
...
     AssetsTreeOperationManagerTest.class,
 
+    // service.db — core coverage
+    DataImportServiceTest.class,
+
     // service.db
     ClientImportEntityResolverTest.class,
 
+    // service.json — core coverage
+    DefaultJsonDataServiceTest.class,
```
etendobot: **Suggestion**

Merge the new test classes into the existing package groups and maintain alphabetical order (e.g., DefaultJsonDataServiceTest should follow AdvancedQueryBuilderTest).

```java
    // service.json
    AdvancedQueryBuilderTest.class,
+   DefaultJsonDataServiceTest.class,
```

**Rationale**: Grouping test classes alphabetically within their package blocks improves readability. Additionally, creating separate 'core coverage' sub-groups for single classes within an existing package group adds unnecessary complexity.

etendobot (changes requested):

src-test/src/org/openbravo/advpaymentmngt/utility/FINUtilityTest.java
```diff
@@ -0,0 +1,271 @@
...
+    }
+  }
+
+  // --- getDaysBetween tests ---
+
+  @Test
+  public void testGetDaysBetweenSameDate() {
+    Date now = new Date();
+    assertEquals(0L, (long) FIN_Utility.getDaysBetween(now, now));
+  }
```
etendobot: **Suggestion**

Remove the redundant explicit cast to (long).

```java
assertEquals(0L, FIN_Utility.getDaysBetween(now, now));
```

**Rationale**: The method 'FIN_Utility.getDaysBetween' returns a Long object. In Java, when passing a Long to a method expecting a primitive long (like JUnit's assertEquals), unboxing occurs automatically. The explicit cast to (long) is redundant and can be removed for cleaner code. This also applies to the other assertions in this diff.

etendobot (changes requested):

etendobot (approved):

ivanrobledofutit (approved):

PR #1050, opened 2026-06-08
Feature ETP-4200: Inherit client from document in line tax insert functions

Gremiger:
## Summary

- `C_INVOICELINETAX_INSERT` and `C_ORDERLINETAX_INSERT` were reading `AD_CLIENT_ID` from `C_TAX` and using it directly as the owner of the inserted line tax row.
- When using system-level taxes (`ad_client_id = '0'`), this produced `c_invoicelinetax` / `c_orderlinetax` rows owned by client `'0'`, causing `OBSecurityException` when the session tried to operate on them.
- Fix: after reading the tax data, do a separate lookup of `AD_CLIENT_ID` from `C_INVOICE` / `C_ORDER` (guarded by a null check), so the inserted row always belongs to the document's client.

## Implementation

```sql
SELECT ISSUMMARY, CASCADE, RATE, BASEAMOUNT, C_TAXBASE_ID
INTO v_IsSummary, v_Cascade, v_Rate, v_BaseAmount, v_C_TaxBase_ID
FROM C_TAX WHERE C_TAX_ID = p_tax_id;
IF (p_invoice_id IS NOT NULL) THEN
  SELECT AD_CLIENT_ID INTO v_Client_ID FROM C_INVOICE WHERE C_INVOICE_ID = p_invoice_id;
END IF;
```

Same pattern applied to `C_ORDERLINETAX_INSERT` using `C_ORDER`.

Both SELECTs are PK lookups — no performance impact regardless of table size.

## Test plan

- [ ] Create an invoice with a system-level tax (`ad_client_id = '0'`) from a non-system client
- [ ] Confirm `c_invoicelinetax` rows are created with the invoice's `ad_client_id`, not `'0'`
- [ ] Complete the invoice — no `OBSecurityException`
- [ ] Create an order with a system-level tax and verify same behavior on `c_orderlinetax`
- [ ] Verify existing per-client taxes still work (result is identical since `tax.client == invoice.client` in that case)

Closes ETP-4200

### DIFF
--- a/src-db/database/model/functions/C_INVOICELINETAX_INSERT.xml
+++ b/src-db/database/model/functions/C_INVOICELINETAX_INSERT.xml
@@ -59,10 +59,11 @@
     Cur_Taxes RECORD;
   BEGIN --BODY
     v_TaxBaseAmount:=0;
-    SELECT ISSUMMARY, CASCADE, RATE, BASEAMOUNT, AD_CLIENT_ID, C_TAXBASE_ID
+    SELECT t.ISSUMMARY, t.CASCADE, t.RATE, t.BASEAMOUNT, i.AD_CLIENT_ID, t.C_TAXBASE_ID
     INTO v_IsSummary, v_Cascade, v_Rate, v_BaseAmount, v_Client_ID, v_C_TaxBase_ID
-    FROM C_TAX
-    WHERE C_TAX_ID = p_tax_id;
+    FROM C_TAX t, C_INVOICE i
+    WHERE t.C_TAX_ID = p_tax_id
+    AND i.C_INVOICE_ID = p_invoice_id;
     IF(v_BaseAmount='LNA' OR v_BaseAmount='LNATAX') THEN
       v_TaxBaseAmount:=p_LineNetAmt;
     ELSIF (v_BaseAmount='TBA' OR v_BaseAmount='TBATAX') THEN
@@ -100,32 +101,26 @@
       IF(p_invoice_id IS NULL OR p_invoiceline_id IS NULL) THEN
          RAISE_APPLICATION_ERROR(-20000, '@20200@');
       END IF;
-      INSERT
-      INTO C_INVOICELINETAX (
-      C_InvoiceLineTax_ID, C_InvoiceLine_ID, C_Invoice_ID, C_Tax_ID, AD_Client_ID, AD_Org_ID,
-      IsActive, Created, CreatedBy, Updated, UpdatedBy,
-      TaxBaseAmt, TaxAmt, Line)
+      INSERT INTO C_INVOICELINETAX (
+        C_InvoiceLineTax_ID, C_InvoiceLine_ID, C_Invoice_ID, C_Tax_ID, AD_Client_ID, AD_Org_ID,
+        IsActive, Created, CreatedBy, Updated, UpdatedBy,
+        TaxBaseAmt, TaxAmt, Line)
       VALUES (
-      GET_UUID(), p_invoiceline_id, p_invoice_id, p_tax_id, v_Client_ID, p_org_id,
-      'Y', now(), p_user_id, now(), p_user_id,
-      v_TaxBaseAmount, ROUND(ROUND(v_TaxBaseAmount, p_StdPrecision) * v_Rate/100, p_StdPrecision), v_LineNo);
+        GET_UUID(), p_invoiceline_id, p_invoice_id, p_tax_id, v_Client_ID, p_org_id,
+        'Y', now(), p_user_id, now(), p_user_id,
+        v_TaxBaseAmount, ROUND(ROUND(v_TaxBaseAmount, p_StdPrecision) * v_Rate/100, p_StdPrecision), v_LineNo);
     ELSE
       FOR Cur_Taxes IN
-        (SELECT C_Tax_ID, RATE
-         FROM C_TAX
-        WHERE Parent_Tax_ID=p_tax_id
-        ORDER BY Line)
+        (SELECT C_Tax_ID, RATE FROM C_TAX WHERE Parent_Tax_ID=p_tax_id ORDER BY Line)
       LOOP
         C_INVOICELINETAX_INSERT(p_org_id, p_invoice_id, p_invoiceline_id, p_user_id, p_originaltax_id, Cur_Taxes.C_Tax_ID, p_LineNetAmt, p_AlternateTaxBaseAmt, p_StdPrecision);
       END LOOP;
     END IF;
     RETURN;
-    --<<FINISH_PROCESS>>
 EXCEPTION
 WHEN OTHERS THEN
   v_ResultStr:= '@ERROR=' || SQLERRM;
-  DBMS_OUTPUT.PUT_LINE(v_ResultStr) ;
-  ROLLBACK;
+  DBMS_OUTPUT.PUT_LINE(v_ResultStr);
   RAISE;
 END C_INVOICELINETAX_INSERT
 ]]></body>
--- a/src-db/database/model/functions/C_INVOICELINETAX_INSERT.xml
+++ b/src-db/database/model/functions/C_INVOICELINETAX_INSERT.xml
@@ -59,11 +59,13 @@
     Cur_Taxes RECORD;
   BEGIN --BODY
     v_TaxBaseAmount:=0;
-    SELECT t.ISSUMMARY, t.CASCADE, t.RATE, t.BASEAMOUNT, i.AD_CLIENT_ID, t.C_TAXBASE_ID
-    INTO v_IsSummary, v_Cascade, v_Rate, v_BaseAmount, v_Client_ID, v_C_TaxBase_ID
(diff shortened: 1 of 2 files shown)


src-db/database/model/functions/C_INVOICELINETAX_INSERT.xml
```diff
@@ -59,10 +59,11 @@
     Cur_Taxes RECORD;
   BEGIN --BODY
     v_TaxBaseAmount:=0;
-    SELECT ISSUMMARY, CASCADE, RATE, BASEAMOUNT, AD_CLIENT_ID, C_TAXBASE_ID
+    SELECT t.ISSUMMARY, t.CASCADE, t.RATE, t.BASEAMOUNT, i.AD_CLIENT_ID, t.C_TAXBASE_ID
     INTO v_IsSummary, v_Cascade, v_Rate, v_BaseAmount, v_Client_ID, v_C_TaxBase_ID
```
etendobot: **[RESOLVED]**  > [!WARNING]
**⚠️ Blocking Issue**

Query the AD_CLIENT_ID from C_INVOICE separately or ensure the p_invoice_id null check happens before this join. Given the recursive nature and existing structure, separate queries or a check-first approach is safer.

```xml
    SELECT ISSUMMARY, CASCADE, RATE, BASEAMOUNT, C_TAXBASE_ID
    INTO v_IsSummary, v_Cascade, v_Rate, v_BaseAmount, v_C_TaxBase_ID
    FROM C_TAX
    WHERE C_TAX_ID = p_tax_id;

    IF (p_invoice_id IS NOT NULL) THEN
      SELECT AD_CLIENT_ID INTO v_Client_ID FROM C_INVOICE WHERE C_INVOICE_ID = p_invoice_id;
    END IF;
```

**Rationale**: Joining C_TAX and C_INVOICE in a single SELECT INTO query before checking if p_invoice_id is NULL will cause the query to return zero rows if the ID is missing. In PL/SQL, this raises a NO_DATA_FOUND exception, preventing the function from reaching the explicit parameter validation check (line 100) that raises the specific '@20200@' application error. This makes the code less robust and harder to debug.

src-db/database/model/functions/C_INVOICELINETAX_INSERT.xml
```diff
@@ -59,10 +59,11 @@
...
-    FROM C_TAX
-    WHERE C_TAX_ID = p_tax_id;
+    FROM C_TAX t, C_INVOICE i
+    WHERE t.C_TAX_ID = p_tax_id
```
etendobot: **Suggestion**

Consider using explicit JOIN syntax for better readability and alignment with modern SQL standards.

```xml
    FROM C_TAX t
    JOIN C_INVOICE i ON i.C_INVOICE_ID = p_invoice_id
```

**Rationale**: While valid in PL/SQL, comma-separated join syntax is considered legacy. Explicit JOIN syntax is clearer and follows modern SQL standards.

src-db/database/model/functions/C_INVOICELINETAX_INSERT.xml
```diff
@@ -59,10 +59,11 @@
     Cur_Taxes RECORD;
   BEGIN --BODY
```
etendobot, line 60: **Suggestion**

Verify if the removal of ROLLBACK is intentional and safe for all calling contexts. In Etendo, it is often better to let the caller handle the transaction, but changing existing behavior should be done with caution.

```xml
  ROLLBACK;
```

**Rationale**: The original code included a ROLLBACK in the exception block. While Etendo's DAL handles transactions for Java-initiated calls, removing it from a PL/SQL helper function that might be called from other database processes can change the expected behavior of sub-transactions. If the intent is to modernize the error handling, ensure that all calling paths (including other PL/SQL procedures) are accounted for.

src-db/database/model/functions/C_ORDERLINETAX_INSERT.xml
```diff
@@ -59,10 +59,11 @@
     Cur_Taxes RECORD;
   BEGIN --BODY
     v_TaxBaseAmount:=0;
-    SELECT ISSUMMARY, CASCADE, RATE, BASEAMOUNT, AD_CLIENT_ID, C_TAXBASE_ID
+    SELECT t.ISSUMMARY, t.CASCADE, t.RATE, t.BASEAMOUNT, o.AD_CLIENT_ID, t.C_TAXBASE_ID
     INTO v_IsSummary, v_Cascade, v_Rate, v_BaseAmount, v_Client_ID, v_C_TaxBase_ID
-    FROM C_TAX
-    WHERE C_TAX_ID = p_tax_id;
+    FROM C_TAX t, C_ORDER o
+    WHERE t.C_TAX_ID = p_tax_id
```
etendobot: **Suggestion**

Use explicit INNER JOIN syntax and consider if the order lookup can be optimized for recursive calls.

```sql
-- Consider selecting from C_ORDER once if possible or keeping lookup separate to avoid redundancy in recursive calls.
    SELECT t.ISSUMMARY, t.CASCADE, t.RATE, t.BASEAMOUNT, o.AD_CLIENT_ID, t.C_TAXBASE_ID
    INTO v_IsSummary, v_Cascade, v_Rate, v_BaseAmount, v_Client_ID, v_C_TaxBase_ID
    FROM C_TAX t
    INNER JOIN C_ORDER o ON (o.C_ORDER_ID = p_order_id)
    WHERE t.C_TAX_ID = p_tax_id;
```

**Rationale**: The current implementation performs a join with C_ORDER in every recursive call of the function. For tax trees with multiple levels, this results in redundant lookups of the same order. While functional, using an explicit INNER JOIN is also preferred for readability over comma-separated tables.

etendobot (changes requested):

src-db/database/model/functions/C_ORDERLINETAX_INSERT.xml
```diff
@@ -59,10 +59,13 @@
     Cur_Taxes RECORD;
   BEGIN --BODY
     v_TaxBaseAmount:=0;
-    SELECT ISSUMMARY, CASCADE, RATE, BASEAMOUNT, AD_CLIENT_ID, C_TAXBASE_ID
-    INTO v_IsSummary, v_Cascade, v_Rate, v_BaseAmount, v_Client_ID, v_C_TaxBase_ID
+    SELECT ISSUMMARY, CASCADE, RATE, BASEAMOUNT, C_TAXBASE_ID
+    INTO v_IsSummary, v_Cascade, v_Rate, v_BaseAmount, v_C_TaxBase_ID
```
etendobot, line 63: **Suggestion**

Consider retaining the table alias 't' for the C_TAX table columns.

```xml
SELECT t.ISSUMMARY, t.CASCADE, t.RATE, t.BASEAMOUNT, t.C_TAXBASE_ID
```

**Rationale**: Using table aliases (e.g., 't.') is a best practice in Etendo SQL functions to prevent column name ambiguity, especially if local variables or future joins are added to the scope.

etendobot (dismissed):

src-db/database/model/functions/C_INVOICELINETAX_INSERT.xml
Gremiger: Applied. Split into two separate SELECTs with an explicit null guard:

```sql
SELECT ISSUMMARY, CASCADE, RATE, BASEAMOUNT, C_TAXBASE_ID
INTO v_IsSummary, v_Cascade, v_Rate, v_BaseAmount, v_C_TaxBase_ID
FROM C_TAX WHERE C_TAX_ID = p_tax_id;
IF (p_invoice_id IS NOT NULL) THEN
  SELECT AD_CLIENT_ID INTO v_Client_ID FROM C_INVOICE WHERE C_INVOICE_ID = p_invoice_id;
END IF;
```

This preserves the original failure path (the `@20200@` check later still fires for the null case in the non-summary branch) while avoiding the NO_DATA_FOUND issue from the join.

src-db/database/model/functions/C_INVOICELINETAX_INSERT.xml
Gremiger: No longer applicable — addressed by splitting into two separate SELECTs (see reply to the blocking comment above). There is no join to refactor.

src-db/database/model/functions/C_INVOICELINETAX_INSERT.xml
Gremiger: Restored. The `ROLLBACK` was accidentally stripped by `export.database` when regenerating the XML from PostgreSQL (PG doesn't support ROLLBACK inside function exception blocks, so the exporter omits it). It's back in the XML now for Oracle compatibility.

src-db/database/model/functions/C_ORDERLINETAX_INSERT.xml
Gremiger: Applied the same split approach as in `C_INVOICELINETAX_INSERT`: two separate SELECTs with a null guard on `p_order_id`. No join remains to refactor.

fran-roig (approved):

PR #1051, opened 2026-06-09
Issue #1049: Fix TypeError when sending emails to multiple customers

### LINKED ISSUE #1049: ETP-4197: Email sending fails with TypeError when selecting invoices from multiple customers
## Error's description

When selecting two or more completed Sales Invoices from **different customers** and clicking the Email icon, the "Email options" popup has multiple issues:

1. **`#toEmail` crashes on load**: The `onLoadDo()` and `validateAndSendEmail()` functions in `send.html` unconditionally access `document.getElementById('toEmail').value` in **3 places**, but `PrintController.getHiddenTags()` removes the `#toEmail` element from the DOM when `moreThanOneCustomer == true`. This causes an uncaught JavaScript TypeError that leaves the Send button non-functional — no emails are sent.
2. **`#toEmail` is not populated with all Business Partners' emails**: In Etendo 25, when multiple customers were selected, the `To` field was auto-populated with the default email of each Business Partner, separated by `;`. This behavior was lost in 26.1.x. The field should always be visible, auto-populated with all default emails (editable by the user before sending), and never hidden.
3. **Send button is not disabled when required fields are missing**: The Send button is active even when required fields (`To`, `Reply-To`, `Subject`, `Template to use`) are incomplete, leading to uncontrolled failures.
4. **Popup always shows new version**: The new popup UI (introduced in ETP-3208) should always be rendered, with the `From` field visible and the `To` field auto-completed and editable — regardless of how many customers are selected.
Exact error in browser console:

```
Uncaught TypeError: Cannot read properties of null (reading 'value')
    at onLoadDo (send.html:193)
    at onload (send.html:573)
```

## Steps to reproduce the error

**Required configuration:**
A valid SMTP configuration must be set up at Client or Organization level (`General Setup > Application > Client > Email Configuration tab`), otherwise the system blocks the flow before the popup opens with a "No sender defined" error.

1. Log in to Etendo with the F&B demo dataset (or any dataset with multiple customers).
2. 2. Navigate to **General Setup > Application > Client**, open the F&B International Group record, go to the **Email Configuration** tab and configure a valid SMTP server (e.g. `smtp.gmail.com`, port `465`, SSL, with authentication).
3. 3. Navigate to the **Sales Invoice** window.
4. 4. Clear the transactional filter to see all invoices.
5. 5. Select two or more **completed** invoices from **different customers** (e.g. doc `1000000` — Restaurantes Luna Llena, S.A. and doc `1000001` — Hoteles Buenas Noches, S.A.).
6. 6. Click the **Email** icon in the toolbar.
7. 7. Observe the "Email options" popup:
8.    - The **`To`** field should be auto-populated with `email_cliente1@example.com; email_cliente2@example.com` but instead it is absent or empty.
9.    - The `Send` button is active despite required fields being missing.
10. 8. Open the browser console (F12): the TypeError is already thrown on popup load.
11. 9. Click **Send** — no action occurs, the error repeats in the console.
## Expected behavior

- The popup should **always** render in its new version (with `From` visible and `To` editable).
- - When multiple customers are selected, the `To` field should be **auto-populated** with the default email of each Business Partner, separated by `;` (e.g. `bp1@example.com; bp2@example.com`). The user should be able to edit the list before sending.
- - - When a single customer is selected, contact selection for that Business Partner is allowed as before.
- - - - The **Send button must be disabled** until all required fields are filled: `To`, `Reply-To`, `Subject`, and `Template to use`.
- - - - - No JavaScript errors should be thrown during popup load or send execution.
## Affected Version

- Etendo version: `26.1.0` → `26.1.8` (latest)
- - Module/Bundle affected: `Etendo Core`
- - - Module version: Regression introduced in release `26.1.0` as part of task `ETP-3208` (2026-03-09).
### Root cause

Three separate issues in `send.html` introduced/regressed in 26.1.x:

1. `onLoadDo()` and `validateAndSendEmail()` access `document.getElementById('toEmail').value` in **3 places** without null checks. `PrintController.getHiddenTags()` (`PrintController.java:1419`) removes `#toEmail` when `moreThanOneCustomer == true`, causing a crash.
2. 2. The multi-customer email aggregation logic from Etendo 25 was not preserved. Previously, the `To` field was populated server-side with all BPs' default emails joined by `;`. This needs to be restored in `PrintController` and reflected in `send.html`.
3. 3. No client-side validation guards the Send button — it should be disabled until `To`, `Reply-To`, `Subject`, and `Template to use` are all filled.
### Proposed fix

**Fix 1 — Null check (3 occurrences in `send.html`):**
```javascript
// Before (broken):
const toEmail = document.getElementById('toEmail').value.trim();

// After (fixed):
const toEmailEl = document.getElementById('toEmail');
const toEmail = toEmailEl ? toEmailEl.value.trim() : '';
```

**Fix 2 — Restore multi-customer email aggregation (`PrintController.java`):**
Instead of removing `#toEmail` when `moreThanOneCustomer == true`, populate it with all BPs' default emails separated by `;`. The field should remain visible and editable.

**Fix 3 — Disable Send button until required fields are complete (`send.html`):**
Add a validation listener that enables the Send button only when `To`, `Reply-To`, `Subject`, and `Template to use` are all non-empty.

**Fix 4 — Always render new popup version:**
Ensure the new popup UI (with `From` visible) is always rendered, regardless of the number of selected customers.

## Other test cases

**Given:** Two or more completed Sales Invoices from different customers are selected and the Email icon is clicked.
**When:** The "Email options" popup opens.
**Then:** The `To` field is auto-populated with all BPs' default emails separated by `;`, the `From` field is visible, and the Send button is disabled until `To`, `Reply-To`, `Subject`, and `Template to use` are all filled.

## Jira issue

https://etendoproject.atlassian.net/browse/ETP-4197

RomanMagnoli:
## Summary

- Fixes uncaught `TypeError: Cannot read properties of null (reading 'value')` thrown in `EmailOptions.html` when selecting invoices from multiple customers and clicking the Email icon.
- Root cause: `PrintController.getHiddenTags()` removes the `<tr id="to">` row (and its child `<input id="toEmail">`) from the DOM when `moreThanOneCustomer == true`, but `onLoadDo()` and `validateAndSendEmail()` accessed `#toEmail.value` unconditionally.
- Two null-guard fixes applied in `EmailOptions.html`:
  1. `onLoadDo()`: guard the BP-contacts AJAX call that reads `toEmail.value` on page load.
  2. `validateAndSendEmail()`: skip the `To` field validation entirely when the element is absent (server already handles per-customer routing in multi-customer mode).

## Test plan

- [ ] Configure a valid SMTP server at Client level (General Setup > Application > Client > Email Configuration).
- [ ] Navigate to Sales Invoice, clear the transactional filter.
- [ ] Select two or more completed invoices from **different customers**.
- [ ] Click the Email icon — popup should open without console errors, showing "This email will be sent to N customers".
- [ ] Click Send — emails should be dispatched successfully.
- [ ] Repeat with invoices from a **single customer** — verify the To field still appears, validates correctly, and Send works as before.

Closes #1049

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/org/openbravo/erpCommon/utility/reporting/printing/EmailOptions.html
+++ b/src/org/openbravo/erpCommon/utility/reporting/printing/EmailOptions.html
@@ -96,6 +96,22 @@
       }
     }
 
+    function validateOptionalEmail(fieldId) {
+      var el = document.getElementById(fieldId);
+      if (!el) { return false; }
+      if (!el.value.trim()) {
+        setEmailError(fieldId, null);
+        return false;
+      }
+      var invalid = findInvalidAddress(el.value);
+      if (invalid !== null) {
+        setEmailError(fieldId, '"' + invalid + '" is not a valid email address.');
+        return true;
+      }
+      setEmailError(fieldId, null);
+      return false;
+    }
+
     function validateAndSendEmail() {
       var hasError = false;
       var toEmailEl = document.getElementById('toEmail');
@@ -114,30 +130,8 @@
           }
         }
       }
-      var ccEl = document.getElementById('ccEmail');
-      if (ccEl && ccEl.value.trim()) {
-        var invalidCC = findInvalidAddress(ccEl.value);
-        if (invalidCC !== null) {
-          setEmailError('ccEmail', '"' + invalidCC + '" is not a valid email address.');
-          hasError = true;
-        } else {
-          setEmailError('ccEmail', null);
-        }
-      } else if (ccEl) {
-        setEmailError('ccEmail', null);
-      }
-      var bccEl = document.getElementById('bccEmail');
-      if (bccEl && bccEl.value.trim()) {
-        var invalidBCC = findInvalidAddress(bccEl.value);
-        if (invalidBCC !== null) {
-          setEmailError('bccEmail', '"' + invalidBCC + '" is not a valid email address.');
-          hasError = true;
-        } else {
-          setEmailError('bccEmail', null);
-        }
-      } else if (bccEl) {
-        setEmailError('bccEmail', null);
-      }
+      if (validateOptionalEmail('ccEmail')) { hasError = true; }
+      if (validateOptionalEmail('bccEmail')) { hasError = true; }
 
       if (!hasError) {
         submitThisPage('EMAIL');
--- a/src/org/openbravo/erpCommon/utility/reporting/printing/EmailOptions.html
+++ b/src/org/openbravo/erpCommon/utility/reporting/printing/EmailOptions.html
@@ -98,17 +98,20 @@
 
     function validateAndSendEmail() {
       var hasError = false;
-      var toVal = document.getElementById('toEmail').value.trim();
-      if (!toVal) {
-        setEmailError('toEmail', 'Recipient email is required.');
-        hasError = true;
-      } else {
-        var invalidTo = findInvalidAddress(toVal);
-        if (invalidTo !== null) {
-          setEmailError('toEmail', '"' + invalidTo + '" is not a valid email address.');
+      var toEmailEl = document.getElementById('toEmail');
+      if (toEmailEl !== null) {
+        var toVal = toEmailEl.value.trim();
+        if (!toVal) {
+          setEmailError('toEmail', 'Recipient email is required.');
           hasError = true;
         } else {
-          setEmailError('toEmail', null);
+          var invalidTo = findInvalidAddress(toVal);
+          if (invalidTo !== null) {
+            setEmailError('toEmail', '"' + invalidTo + '" is not a valid email address.');
+            hasError = true;
+          } else {
+            setEmailError('toEmail', null);
+          }
         }
       }
       var ccEl = document.getElementById('ccEmail');
@@ -217,7 +220,8 @@
         setWindowElementFocus('firstElement');
         resizeWindow();
 
-        if (document.getElementById('toEmail').value === '') {
+        var toEmailEl = document.getElementById('toEmail');
+        if (toEmailEl && toEmailEl.value === '') {
           var warnCallback = function(paramXMLParticular, XMLHttpRequestObj) {
             if (getReadyStateHandler(XMLHttpRequestObj)) {
               try {


src/org/openbravo/erpCommon/utility/reporting/printing/EmailOptions.html
```diff
@@ -98,17 +98,20 @@
...
-      if (!toVal) {
-        setEmailError('toEmail', 'Recipient email is required.');
-        hasError = true;
-      } else {
-        var invalidTo = findInvalidAddress(toVal);
-        if (invalidTo !== null) {
-          setEmailError('toEmail', '"' + invalidTo + '" is not a valid email address.');
+      var toEmailEl = document.getElementById('toEmail');
+      if (toEmailEl !== null) {
+        var toVal = toEmailEl.value.trim();
```
etendobot, line 119: **Suggestion**

Use a consistent truthy check for the DOM element.

```javascript
if (toEmailEl) {
```

**Rationale**: While `!== null` works, a truthy check `if (toEmailEl)` is more concise and idiomatic in JavaScript for DOM elements, as `document.getElementById` returns `null` (falsy) when not found. Also, later in the same file (line 224), a truthy check is used.

etendobot (approved):

etendobot (approved):

etendobot (approved):

src/org/openbravo/erpCommon/utility/reporting/printing/EmailOptions.html
```diff
@@ -96,45 +96,42 @@
...
+            hasError = true;
+          } else {
+            setEmailError('toEmail', null);
+          }
         }
-      } else if (bccEl) {
-        setEmailError('bccEmail', null);
       }
+      if (validateOptionalEmail('ccEmail')) { hasError = true; }
+      if (validateOptionalEmail('bccEmail')) { hasError = true; }
```
etendobot, line 134: **Suggestion**

Consider using a logical OR assignment to update the error state.

```javascript
hasError = validateOptionalEmail('ccEmail') || hasError;
```

**Rationale**: While the current implementation is functional, using a logical OR assignment is a more idiomatic way to track error states in JavaScript loops or sequences.

src/org/openbravo/erpCommon/utility/reporting/printing/EmailOptions.html
```diff
@@ -96,45 +96,42 @@
...
-          setEmailError('toEmail', '"' + invalidTo + '" is not a valid email address.');
-          hasError = true;
-        } else {
-          setEmailError('toEmail', null);
-        }
+    function validateOptionalEmail(fieldId) {
+      var el = document.getElementById(fieldId);
+      if (!el) { return false; }
+      if (!el.value.trim()) {
+        setEmailError(fieldId, null);
```
etendobot, line 103: **Suggestion**

Add a defensive check to ensure el.value is not null/undefined before trimming.

```javascript
if (!el.value || !el.value.trim()) {
```

**Rationale**: Although HTMLInputElement.value is typically a string, adding a defensive check for the existence of the value before calling trim() prevents potential TypeErrors in edge cases.

etendobot (approved):

Agustin-Calderon (approved):

jortolanosmf (approved):

RubenEtendo (approved):

ivanrobledofutit (approved):

src/org/openbravo/erpCommon/utility/reporting/printing/EmailOptions.html
```diff
@@ -96,45 +96,42 @@
...
-      } else {
-        var invalidTo = findInvalidAddress(toVal);
-        if (invalidTo !== null) {
-          setEmailError('toEmail', '"' + invalidTo + '" is not a valid email address.');
-          hasError = true;
-        } else {
-          setEmailError('toEmail', null);
-        }
+    function validateOptionalEmail(fieldId) {
+      var el = document.getElementById(fieldId);
```
etendobot, line 100: **Suggestion**

The new 'validateOptionalEmail' function is a good refactor. Consider adding a defensive check for 'el.value' type to ensure 'trim()' is only called on strings.

```javascript
    function validateOptionalEmail(fieldId) {
      var el = document.getElementById(fieldId);
      if (!el || typeof el.value !== 'string') { return false; }
      if (!el.value.trim()) {
        setEmailError(fieldId, null);
        return false;
      }
      var invalid = findInvalidAddress(el.value);
      if (invalid !== null) {
        setEmailError(fieldId, '"' + invalid + '" is not a valid email address.');
        return true;
      }
      setEmailError(fieldId, null);
      return false;
    }
```

**Rationale**: Introducing a helper function to handle optional email validation reduces code duplication and improves maintainability. Adding a check for 'typeof el.value === "string"' ensures that the function won't throw a TypeError if 'document.getElementById' returns an element that is not a form field (e.g., if the ID is reused or the UI structure changes).

src/org/openbravo/erpCommon/utility/reporting/printing/EmailOptions.html
etendobot, line 134: **Suggestion**

Consider using a more concise assignment for updating the 'hasError' flag.

```javascript
      hasError = validateOptionalEmail('ccEmail') || hasError;
      hasError = validateOptionalEmail('bccEmail') || hasError;
```

**Rationale**: Using the results of the validation function directly to update 'hasError' is cleaner than using 'if' statements. However, the current implementation is correct and functional.

etendobot (approved):

Issue #1053, opened 2026-06-11
ETP-4234: QtyDelivered not reverted on Goods Receipt void, causing MRP inconsistencies and blocking new receipts on Purchase Orders


juanfunes1996:
# Error's description
When a Goods Receipt linked to a Purchase Order is voided/reactivated, the QtyDelivered field on the Purchase Order line is not decremented. This causes two impacts:

1. Attempting to create a new Goods Receipt against the same PO line fails with: "Movement Quantity is greater than the Ordered Quantity"
2. The MRP (Process Purchase Plan) excludes the PO line from the plan because it interprets it as fully received, making pending units invisible to procurement planning.

 # Steps to reproduce the error
Flow 1 — Goods Receipt / Purchase Order:
1. Create and book a Purchase Order with 10 units of any stocked product.
2. Create a Goods Receipt from that PO line and complete it → QtyDelivered updates to 10.
3. Void/reactivate the Goods Receipt.
4. Verify QtyDelivered on the PO line — it remains at 10 instead of reverting to 0.
5. Try to create a new Goods Receipt from the same PO line → error: "Movement Quantity is greater than the Ordered Quantity".

https://github.com/user-attachments/assets/dc6398dd-169c-4c6a-ad60-3b8f56856a70

Flow 2 — Purchasing Plan / MRP:
1. Create and book a Purchase Order with 10 units.
2. Create a Purchasing Plan for that product and run Process Purchase Plan → 10 units appear as pending supply.
3. Create and complete a Goods Receipt from the PO → QtyDelivered updates to 10.
4. Delete the plan lines and run Process Purchase Plan again → quantities appear grouped/inflated with other lines of the same product.
5. Delete all plan lines and void the Goods Receipt → QtyDelivered stays at 10.
6. Run Process Purchase Plan again → the 10 units disappear from the plan.

https://github.com/user-attachments/assets/12cf7f80-6d5b-4be8-b86d-d7fd485e5981

 # Expected behavior
Voiding a Goods Receipt should decrement QtyDelivered on the Purchase Order line back to its previous value.
The MRP should continue showing the PO line as pending supply after a void.
New Goods Receipts should be creatable against the same PO line after voiding a prior one.

 # Affected Version
Etendo version: 25.4.17 | 26.1.9
Module/Bundle affected: Etendo Core

 # Solution Design (optional)
**Root cause**
[...]

PR #1057, opened 2026-06-16
Issue #1053: Fix PO QtyDelivered on receipt void and MRP double-count

fran-roig:
ETP-4234

### DIFF
--- a/src-db/database/model/functions/M_INOUT_POST.xml
+++ b/src-db/database/model/functions/M_INOUT_POST.xml
@@ -880,7 +880,7 @@
                   IF (v_DocStatus = 'CO') THEN
                     UPDATE C_ORDERLINE
                       SET QtyDelivered = QtyDelivered + v_QtyPO,
-                      QtyReserved = QtyReserved + v_QtyPO,
+                      QtyReserved = QtyReserved - v_QtyPO,
                       DATEDELIVERED = (CASE WHEN ABS(QtyDelivered + v_QtyPO) > 0
                                       THEN v_DateDelivered ELSE NULL END),
                       Updated = now(),
@@ -889,7 +889,7 @@
                   ELSE 
                     UPDATE C_ORDERLINE
                       SET QtyDelivered = QtyDelivered + v_QtyPO,
-                      QtyReserved = QtyReserved + v_QtyPO,
+                      QtyReserved = QtyReserved - v_QtyPO,
                       DATEDELIVERED = (CASE WHEN ABS(QtyDelivered + v_QtyPO) > 0
                                       THEN v_DateDelivered ELSE NULL END),
                       Updated = now(),
@@ -915,7 +915,7 @@
                 ELSE
                   UPDATE C_ORDERLINE
                     SET QtyDelivered = QtyDelivered + v_QtyPO,
-                    QtyReserved = QtyReserved + v_QtyPO,
+                    QtyReserved = QtyReserved - v_QtyPO,
                     DATEDELIVERED = (CASE WHEN ABS(QtyDelivered + v_QtyPO) > 0
                                     THEN v_DateDelivered ELSE NULL END),
                     Updated = now(),
--- a/src-db/database/model/functions/M_INOUT_POST.xml
+++ b/src-db/database/model/functions/M_INOUT_POST.xml
@@ -879,18 +879,18 @@
                 ELSE
                   IF (v_DocStatus = 'CO') THEN
                     UPDATE C_ORDERLINE
-                      SET QtyDelivered = QtyDelivered + LEAST(ABS(v_QtyPO), (ABS(QtyOrdered) - ABS(QtyDelivered))),
-                      QtyReserved = QtyReserved + LEAST(ABS(v_QtyPO), (ABS(QtyOrdered) - ABS(QtyReserved))),
-                      DATEDELIVERED = (CASE WHEN (ABS(QtyDelivered) + LEAST(ABS(v_QtyPO), (ABS(QtyOrdered) - ABS(QtyDelivered)))) > 0 
+                      SET QtyDelivered = QtyDelivered + (CASE WHEN v_QtyPO < 0 THEN v_QtyPO ELSE LEAST(v_QtyPO, (ABS(QtyOrdered) - ABS(QtyDelivered))) END),
+                      QtyReserved = QtyReserved + (CASE WHEN v_QtyPO < 0 THEN v_QtyPO ELSE LEAST(v_QtyPO, (ABS(QtyOrdered) - ABS(QtyReserved))) END),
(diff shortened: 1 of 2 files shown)


src-db/database/model/functions/M_INOUT_POST.xml
```diff
@@ -879,18 +879,18 @@
                 ELSE
                   IF (v_DocStatus = 'CO') THEN
                     UPDATE C_ORDERLINE
-                      SET QtyDelivered = QtyDelivered + LEAST(ABS(v_QtyPO), (ABS(QtyOrdered) - ABS(QtyDelivered))),
-                      QtyReserved = QtyReserved + LEAST(ABS(v_QtyPO), (ABS(QtyOrdered) - ABS(QtyReserved))),
-                      DATEDELIVERED = (CASE WHEN (ABS(QtyDelivered) + LEAST(ABS(v_QtyPO), (ABS(QtyOrdered) - ABS(QtyDelivered)))) > 0 
+                      SET QtyDelivered = QtyDelivered + (CASE WHEN v_QtyPO < 0 THEN v_QtyPO ELSE LEAST(v_QtyPO, (ABS(QtyOrdered) - ABS(QtyDelivered))) END),
+                      QtyReserved = QtyReserved + (CASE WHEN v_QtyPO < 0 THEN v_QtyPO ELSE LEAST(v_QtyPO, (ABS(QtyOrdered) - ABS(QtyReserved))) END),
```
etendobot: **[RESOLVED]**  > [!WARNING]
**⚠️ Blocking Issue**

Implement a sign-aware logic that distinguishes between receiving (signs of v_QtyPO and QtyOrdered match) and voiding (signs differ). Use the capped absolute value only when receiving.

```sql
SET QtyDelivered = QtyDelivered + (CASE WHEN (v_QtyPO < 0 AND QtyOrdered >= 0) OR (v_QtyPO > 0 AND QtyOrdered < 0) THEN v_QtyPO ELSE (CASE WHEN v_QtyPO < 0 THEN -1 ELSE 1 END) * LEAST(ABS(v_QtyPO), ABS(QtyOrdered) - ABS(QtyDelivered)) END),
```

**Rationale**: The new logic incorrectly assumes that voiding always involves a negative quantity and receiving always involves a positive one. In Return Material Receipts (RMAs), receiving a return uses a negative quantity, and voiding that return uses a positive quantity. The proposed 'CASE' logic will: 1) Fail to revert quantities when voiding a return (it will add 0). 2) Allow over-delivery on returns because 'LEAST' between a negative quantity and a positive remaining balance will pick the negative quantity.

src-db/database/model/functions/M_INOUT_POST.xml
```diff
@@ -879,18 +879,18 @@
...
+                      DATEDELIVERED = (CASE WHEN (QtyDelivered + (CASE WHEN v_QtyPO < 0 THEN v_QtyPO ELSE LEAST(v_QtyPO, (ABS(QtyOrdered) - ABS(QtyDelivered))) END)) > 0
                                       THEN v_DateDelivered ELSE NULL END),
```
etendobot, line 885: **[RESOLVED]**  > [!WARNING]
**⚠️ Blocking Issue**

Re-introduce 'ABS()' in the condition check to ensure the delivery date is correctly maintained for both standard and return orders.

```sql
DATEDELIVERED = (CASE WHEN ABS(QtyDelivered + (CASE WHEN (v_QtyPO < 0 AND QtyOrdered >= 0) OR (v_QtyPO > 0 AND QtyOrdered < 0) THEN v_QtyPO ELSE (CASE WHEN v_QtyPO < 0 THEN -1 ELSE 1 END) * LEAST(ABS(v_QtyPO), ABS(QtyOrdered) - ABS(QtyDelivered)) END)) > 0
```

**Rationale**: Removing 'ABS()' from the condition check breaks return orders. For returns, the 'QtyDelivered' is negative, so the condition '(QtyDelivered + ...) > 0' will always be false, causing 'DATEDELIVERED' to be cleared (NULL) even when the return is successfully received.

etendobot (changes requested):

src-db/database/model/functions/M_INOUT_POST.xml
etendobot: **[RESOLVED]**  > [!WARNING]
**⚠️ Blocking Issue**

Implement the sign-aware logic that distinguishes between receiving (signs of v_QtyPO and QtyOrdered match) and voiding (signs differ). Use the capped absolute value only when receiving.

```xml
SET QtyDelivered = QtyDelivered + (CASE WHEN (v_QtyPO < 0 AND QtyOrdered >= 0) OR (v_QtyPO > 0 AND QtyOrdered < 0) THEN v_QtyPO ELSE (CASE WHEN v_QtyPO < 0 THEN -1 ELSE 1 END) * LEAST(ABS(v_QtyPO), ABS(QtyOrdered) - ABS(QtyDelivered)) END),
```

**Rationale**: The logic 'v_QtyPO < 0' is insufficient. When receiving a return (RMA), 'v_QtyPO' is negative and 'QtyOrdered' is also negative. In this case, the 'LEAST' capping logic must still be applied to prevent receiving more than what was ordered. The provided logic bypasses the cap for all negative quantities, which is only correct when voiding a standard receipt, not when receiving a return.

src-db/database/model/functions/M_INOUT_POST.xml
etendobot, line 885: **[RESOLVED]**  > [!WARNING]
**⚠️ Blocking Issue**

Re-introduce 'ABS()' in the condition check to ensure the delivery date is correctly maintained for both standard and return orders.

```xml
DATEDELIVERED = (CASE WHEN ABS(QtyDelivered + (CASE WHEN (v_QtyPO < 0 AND QtyOrdered >= 0) OR (v_QtyPO > 0 AND QtyOrdered < 0) THEN v_QtyPO ELSE (CASE WHEN v_QtyPO < 0 THEN -1 ELSE 1 END) * LEAST(ABS(v_QtyPO), ABS(QtyOrdered) - ABS(QtyDelivered)) END)) > 0
```

**Rationale**: For Return Material Receipts, 'QtyDelivered' is a negative value. The condition '(QtyDelivered + ...) > 0' will evaluate to false for returns (e.g., -10 > 0 is false), causing the 'DATEDELIVERED' to be cleared (set to NULL) even when the return is successfully received. Using 'ABS(...) > 0' ensures the logic works for both standard (positive) and return (negative) quantities.

etendobot (changes requested):

etendobot (approved):

RomanMagnoli (approved):

ivanrobledofutit (approved):

PR #1085, opened 2026-07-13
Feature ETP-4469: Fix transactional sequence stale-entity race

sebastianbarrozo:
## Summary

Fixes the root cause of [ETP-4469](https://etendoproject.atlassian.net/browse/ETP-4469): under concurrent transactions, two callers of the transactional sequence engine could compute the **same** next value (observed in production as duplicate `EM_Etgo_Identifier` -> `c_bpartner_value` unique-constraint violations during concurrent Contacts CSV `/batch` imports).

## Root cause

`DefaultTransactionalSequence.generateValue()` first looks the `Sequence` entity up **unlocked** (`getSequenceFromParameters()`), hydrating it into the Hibernate session. `TransactionalSequenceUtils.lockSequence()` then runs a locking HQL query - but Hibernate resolves the query result against the entity instance already cached in the session (identity map) and **discards the fresh row values**. A transaction that preloaded the sequence while another transaction held the lock therefore computes its value from the stale `nextAssignedNumber` and produces a duplicate, even though the SQL-level `FOR UPDATE` worked.

## Fix

`lockSequence()` now uses `session.flush()` + `session.refresh(sequence, LockOptions.UPGRADE)`:
- `refresh(..., UPGRADE)` acquires the `SELECT ... FOR UPDATE` lock **and** rehydrates the entity from the locked row, atomically.
- The preceding `flush()` preserves increments pending in the current session (several values generated inside one transaction), which a refresh would otherwise discard.

## Tests

- **New** `DefaultTransactionalSequencePreloadRaceTest`: reproduces the production interleaving (T1 computes and holds its transaction open, T2 preloads unlocked during that window, T1 commits, T2 asks for the locked value). RED before the fix (both threads got `1000000`), GREEN after (`1000000` / `1000001`).
- No regression: `TransactionalSequenceUtilsTest` 3/3, `TransactionalSequenceUpdateNextTest` 2/2, Spock `nextvaluegeneration.TransactionalSequenceUtilsTest` (16-case concurrency table) 16/16, `SequenceUtilsTest` 10/10.

## Related

Companion branch `feature/ETP-4469` in `com.etendoerp.go` adds a defense-in-depth savepoint in `BusinessPartnerHandler.afterHandle()` so any future duplicate-key hit no longer poisons the whole `/batch` transaction, plus `@Vetoed` on handler test doubles so Weld-based tests run without CDI ambiguity.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

[ETP-4469]: https://etendoproject.atlassian.net/browse/ETP-4469?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src-test/src/com/etendoerp/sequences/transactional/BusinessPartnerSequenceGenerationRecursionTest.java
+++ b/src-test/src/com/etendoerp/sequences/transactional/BusinessPartnerSequenceGenerationRecursionTest.java
@@ -1,0 +1,61 @@
+package com.etendoerp.sequences.transactional;
+
+import static org.junit.Assert.assertNotNull;
+
+import java.util.UUID;
+
+import org.junit.Test;
+import org.openbravo.base.provider.OBProvider;
+import org.openbravo.dal.service.OBDal;
+import org.openbravo.model.ad.datamodel.Column;
+import org.openbravo.model.ad.domain.DimensionsList;
+import org.openbravo.model.ad.domain.SequenceConfig;
+import org.openbravo.model.ad.utility.Sequence;
+import org.openbravo.model.common.businesspartner.BusinessPartner;
+import org.openbravo.model.common.businesspartner.Category;
+import org.openbravo.test.base.OBBaseTest;
+
+/**
+ * Reproduces a {@link StackOverflowError} caused by calling {@code session.flush()} inside
+ * {@link TransactionalSequenceUtils#lockSequence(String)}.
+ *
+ * <p>{@code DefaultTransactionalSequence.generateValue()} is invoked BY Hibernate itself, from
+ * inside {@code AbstractEntityPersister.insert()}, as part of an ALREADY IN-PROGRESS flush
+ * ({@code preInsertInMemoryValueGeneration}, called while {@code ActionQueue.executeActions()}
+ * is iterating). Calling {@code session.flush()} again from {@code lockSequence()} re-enters
+ * that same action-queue processing, which re-triggers {@code generateValue()} for the same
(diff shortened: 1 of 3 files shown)


etendobot (approved):

etendobot (approved):

sebastianbarrozo:
**Update:** the first version of this fix (commit 9695f5f3) introduced a critical regression — `session.flush()` inside `lockSequence()` recurses infinitely (`StackOverflowError`) because `generateValue()` is invoked by Hibernate itself from inside an already-in-progress flush (`preInsertInMemoryValueGeneration`, triggered by any `BusinessPartner` insert). A nested `flush()` call re-enters the same action-queue processing and re-triggers `generateValue()` for the same pending insert.

Fixed in b62c115d by dropping the `flush()` call — `session.refresh(sequence, LockOptions.UPGRADE)` alone is enough, since any pending increment in the current transaction is already flushed either by the outer, already-running flush or by the caller's normal per-record flush (e.g. one row of a batch import).

New regression test `BusinessPartnerSequenceGenerationRecursionTest` saves a real `BusinessPartner` (triggering the actual Hibernate-driven generator path) to catch this — RED with the flush(), GREEN without it. All 33 sequence tests green (unit + Spock concurrency table + this new one).

AyelenGarcia01 (approved):

Gremiger (approved):

RomanMagnoli (approved):

PR #1101, opened 2026-07-24
Feature ETP-4603: Add Stored columns to Etendo Go

valenvivaldi:


epl-1807/NOTES.md
isaiasb-etendo, line 1: todo esto es necesario comitearlo?

src-db/database/model/tables/AD_COLUMN.xml
```diff
@@ -225,6 +225,22 @@
...
+        <default/>
+        <onCreateDefault/>
+      </column>
+      <column name="COMPUTATION_SEQUENCE_NUMBER" primaryKey="false" required="false" type="DECIMAL" size="10,0" autoIncrement="false">
+        <default><![CDATA[0]]></default>
+        <onCreateDefault><![CDATA[0]]></onCreateDefault>
+      </column>
+      <column name="COMPUTATION_MODE" primaryKey="false" required="true" type="VARCHAR" size="60" autoIncrement="false">
+        <default><![CDATA[N]]></default>
+        <onCreateDefault><![CDATA[CASE WHEN SQLLOGIC IS NULL THEN 'N' ELSE 'V' END]]></onCreateDefault>
```
Matias-Bernal: 'V' no sera 'Y'?

sebastianbarrozo (approved):

src-db/database/model/tables/AD_COLUMN_COMP_DEPENDENCY.xml
Matias-Bernal, line 1: Add a trigger for this table to checking that the module is in development are missing.

src-db/database/model/tables/AD_COMPDEP_WATCHED_COL.xml
Matias-Bernal, line 1: Add a trigger for this table to checking that the module is in development are missing.

src-db/database/model/tables/AD_STOREDCOLUMN_DIRTY.xml
Matias-Bernal, line 1: Add a trigger for this table to checking that the module is in development are missing.

Matias-Bernal: AD_MODULE_ID is not requiered here?

src-db/database/model/tables/AD_COMPDEP_WATCHED_COL.xml
Matias-Bernal: AD_MODULE_ID is not requiered here?

src-db/database/sourcedata/AD_DATASET_TABLE.xml
```diff
@@ -741,6 +741,20 @@
...
 
+<!--370BB33949304E0397E1AE869467DD4B--><AD_DATASET_TABLE>
+<!--370BB33949304E0397E1AE869467DD4B-->  <AD_DATASET_TABLE_ID><![CDATA[370BB33949304E0397E1AE869467DD4B]]></AD_DATASET_TABLE_ID>
+<!--370BB33949304E0397E1AE869467DD4B-->  <AD_CLIENT_ID><![CDATA[0]]></AD_CLIENT_ID>
+<!--370BB33949304E0397E1AE869467DD4B-->  <AD_ORG_ID><![CDATA[0]]></AD_ORG_ID>
+<!--370BB33949304E0397E1AE869467DD4B-->  <ISACTIVE><![CDATA[Y]]></ISACTIVE>
+<!--370BB33949304E0397E1AE869467DD4B-->  <AD_DATASET_ID><![CDATA[0]]></AD_DATASET_ID>
+<!--370BB33949304E0397E1AE869467DD4B-->  <AD_TABLE_ID><![CDATA[680EB01827274531AF82CE8094A8E5B5]]></AD_TABLE_ID>
+<!--370BB33949304E0397E1AE869467DD4B-->  <INCLUDEALLCOLUMNS><![CDATA[Y]]></INCLUDEALLCOLUMNS>
+<!--370BB33949304E0397E1AE869467DD4B-->  <WHERECLAUSE><![CDATA[ad_column_comp_dependency_id in (select ad_column_comp_dependency_id from ad_column_comp_dependency where ad_module_id in (:moduleid))]]></WHERECLAUSE>
```
Matias-Bernal: If this table will have AD_MODULE_ID this query is more easy.

src-db/database/sourcedata/AD_REFERENCE.xml
```diff
@@ -4845,6 +4845,19 @@ There is no any kind of timezone conversion. The stored time does not have timez
 <!--2FE25E4AF395479EB8BAB0539A829C28-->  <ISVALUEDISPLAYED><![CDATA[N]]></ISVALUEDISPLAYED>
 <!--2FE25E4AF395479EB8BAB0539A829C28--></AD_REFERENCE>
 
+<!--30D54C1A38A748D79DCC79AB67975639--><AD_REFERENCE>
+<!--30D54C1A38A748D79DCC79AB67975639-->  <AD_REFERENCE_ID><![CDATA[30D54C1A38A748D79DCC79AB67975639]]></AD_REFERENCE_ID>
+<!--30D54C1A38A748D79DCC79AB67975639-->  <AD_CLIENT_ID><![CDATA[0]]></AD_CLIENT_ID>
+<!--30D54C1A38A748D79DCC79AB67975639-->  <AD_ORG_ID><![CDATA[0]]></AD_ORG_ID>
+<!--30D54C1A38A748D79DCC79AB67975639-->  <ISACTIVE><![CDATA[Y]]></ISACTIVE>
+<!--30D54C1A38A748D79DCC79AB67975639-->  <NAME><![CDATA[Computation Mode]]></NAME>
+<!--30D54C1A38A748D79DCC79AB67975639-->  <DESCRIPTION><![CDATA[Mode of computation for a column: Normal, Virtual (SQLLOGIC), or Stored Computed]]></DESCRIPTION>
```
Matias-Bernal, line 4854: A default preference with normal value?

src-db/database/sourcedata/AD_REFERENCE.xml
```diff
@@ -6860,6 +6873,19 @@ By default the type is empty (null), which means the product characteristic does
 <!--CA331CAC476B4B4B82DB4B9B61378C83-->  <ISVALUEDISPLAYED><![CDATA[N]]></ISVALUEDISPLAYED>
 <!--CA331CAC476B4B4B82DB4B9B61378C83--></AD_REFERENCE>
 
+<!--CA87ED57D8BE499B924F8F8ABF00A01D--><AD_REFERENCE>
+<!--CA87ED57D8BE499B924F8F8ABF00A01D-->  <AD_REFERENCE_ID><![CDATA[CA87ED57D8BE499B924F8F8ABF00A01D]]></AD_REFERENCE_ID>
+<!--CA87ED57D8BE499B924F8F8ABF00A01D-->  <AD_CLIENT_ID><![CDATA[0]]></AD_CLIENT_ID>
+<!--CA87ED57D8BE499B924F8F8ABF00A01D-->  <AD_ORG_ID><![CDATA[0]]></AD_ORG_ID>
+<!--CA87ED57D8BE499B924F8F8ABF00A01D-->  <ISACTIVE><![CDATA[Y]]></ISACTIVE>
+<!--CA87ED57D8BE499B924F8F8ABF00A01D-->  <NAME><![CDATA[Refresh Mode]]></NAME>
+<!--CA87ED57D8BE499B924F8F8ABF00A01D-->  <DESCRIPTION><![CDATA[When the stored computed column value is refreshed: Synchronous, Queued async, or Manual]]></DESCRIPTION>
```
Matias-Bernal, line 6882: Preference with default value?

src-test/src/org/openbravo/modulescript/StoredComputedValidatorLiveDbTest.java
```diff
@@ -0,0 +1,587 @@
...
+ * on the test classpath.
+ */
+public class StoredComputedValidatorLiveDbTest extends OBBaseTest {
+
+  private DalConnectionProvider cp;
+  private Connection con;
+  /** Function names created in-transaction; dropped implicitly by the rollback. */
+  private String validFn;
+
+  @Before
```
Matias-Bernal, line 100: Ensure Junit 5 compatibility

src/org/openbravo/base/gen/entity.ftl
```diff
@@ -131,7 +131,7 @@ public class ${entity.simpleClassName} extends BaseOBObject ${entity.implementsS
     </#if>
     }
     </#if>
-    <#if !(p.domainType?? && p.domainType.reference.id?matches(SEARCH_VECTOR_REF_ID))>
```
Matias-Bernal, line 134: `!(p.domainType?? && !p.storedComputed && p.domainType.reference.id?matches(SEARCH_VECTOR_REF_ID))>`

src/org/openbravo/dal/core/DalMappingGenerator.java
```diff
@@ -303,7 +303,8 @@ private String generatePrimitiveMapping(Property p) {
     // ignoring p.isUpdatable() for now as this is primarily used
     // for ui and not for background processes
 
-    if (p.isInactive() || p.getEntity().isView() || p.getSqlLogic() != null) {
```
Matias-Bernal, line 306: `if (p.isInactive() || p.isStoredComputed() ||  p.getEntity().isView() || p.getSqlLogic() != null) {`

src-db/database/model/tables/AD_COLUMN.xml
valenvivaldi: `COMPUTATION_MODE` is not a Y/N boolean flag — it is a computation-mode enum. Possible values:

- **`N`** — None: a regular column, not computed (the default; see `Property.java` `computationMode = "N"`).
- **`V`** — Virtual: value derived on read from `SQLLogic`, not materialized (a preexisting virtual column). Used e.g. in `AD_FIELD` display logic `@Computation_Mode@='V'`.
- **`S`** — Stored computed: a physical column materialized and maintained by the engine (this feature). All validators check `"S".equals(computationMode)` / `Property.isStoredComputed()`.
- **`Q`** / **`M`** — other stored/refresh modes (see `STORED-COMPUTED-COLUMNS.md`).

So `'V'` is correct here: the `onCreateDefault` `CASE WHEN SQLLOGIC IS NULL THEN 'N' ELSE 'V' END` seeds any preexisting column that has `SQLLogic` as **Virtual** (`'V'`), and columns without it as **None** (`'N'`). `'Y'` is not a valid value for this field and would break the `'S'`/`'V'`/`'N'` comparisons across the codebase.

src-db/database/model/tables/AD_COMPDEP_WATCHED_COL.xml
valenvivaldi: Good catch. Currently AD_COMPDEP_WATCHED_COL has no AD_MODULE_ID of its own: module ownership is derived from its parent (AD_COLUMN_COMP_DEPENDENCY) via the FK, and the dataset export joins through the parent (see AD_DATASET_TABLE:752). Since it is configuration shipped by a module, we'll add AD_MODULE_ID here so the export simplifies and the dev-in-development trigger can apply directly.

src-db/database/model/tables/AD_STOREDCOLUMN_DIRTY.xml
valenvivaldi: For AD_STOREDCOLUMN_DIRTY it is intentional to not have AD_MODULE_ID. This is a runtime dirty-queue table (TARGET_RECORD_ID, TRANSACTION_ID, RETRY_COUNT, ERROR_MSG, IS_IGNORED, REFRESH_MODE) that the recompute engine fills and drains at runtime. No module ever ships rows here, so it is not part of any dataset and must not carry AD_MODULE_ID nor a module-in-development trigger.

src-test/src/org/openbravo/modulescript/StoredComputedValidatorLiveDbTest.java
valenvivaldi: This test `extends OBBaseTest`, which is JUnit 4 (`@Rule`/`@Before` from `org.junit`), so it follows the JUnit 4 lifecycle like every other `OBBaseTest`-based integration test in the repo. It runs under the JUnit Platform through the vintage engine (already on the test runtime classpath — other JUnit 4 tests in `src-test` execute), so it is not silently skipped. Migrating it to Jupiter would mean dropping `OBBaseTest`'s lifecycle/rollback infrastructure. The deterministic rule matrix is already covered in pure JUnit 5 by the sibling `StoredComputedValidatorRulesTest`; this suite is the best-effort live-DB confirmation on top of the JUnit 4 base. Keeping it on JUnit 4 is intentional.

src/org/openbravo/base/gen/entity.ftl
valenvivaldi: The current guard is intentional: the getter is generated for stored-computed properties (line 113), the setter is not (line 134). Moving `!p.storedComputed` inside the negated group as suggested changes the logic from `!(G) && !S` to `!(G && !S)` = `!G || S`, which would generate setters again for stored-computed columns — exactly what we want to avoid, since a stored-computed column is engine-maintained and must not be writable from Java. Keeping `... && !p.storedComputed` outside the negation preserves getter-yes / setter-no. (G = is a search-vector reference, S = storedComputed.)

src/org/openbravo/dal/core/DalMappingGenerator.java
valenvivaldi: `p.isStoredComputed()` is already part of this OR (line 307), so stored-computed properties already get `insert="false" update="false"`. The suggested edit only reorders the `||` operands; since `||` is commutative and none of these accessors have side effects, it is functionally identical.

Reordering for short-circuit would only help if the later operands were expensive (I/O, heavy computation) or this ran in a hot path — neither applies: all four are trivial in-memory getters on an already-loaded `Property`, evaluated once per property at mapping-generation time. So the reorder would not add anything performance-wise either. Already covered.

src-db/database/model/tables/AD_COLUMN_COMP_DEPENDENCY.xml
valenvivaldi: Added. Created the `AD_COLCOMP_DEP_MOD_TRG` trigger on `AD_COLUMN_COMP_DEPENDENCY` (before insert/update/delete, for each row) following the standard dev-module protection pattern: it raises `@ChangeNotInDevModule@` when moving the record to a non-dev module, and `@20532@` / `@20533@` when modifying or inserting/deleting from a module that is not in development.

src-db/database/model/tables/AD_COMPDEP_WATCHED_COL.xml
valenvivaldi: Added. Created the `AD_COMPDEP_WCOL_MOD_TRG` trigger on `AD_COMPDEP_WATCHED_COL` with the same dev-module protection pattern. As discussed in this thread, `AD_MODULE_ID` (required) was also added to the table, so the trigger checks module ownership directly instead of deriving it from the parent — and the dataset export whereclause was simplified accordingly (see `AD_DATASET_TABLE:752`).

src-db/database/sourcedata/AD_DATASET_TABLE.xml
valenvivaldi: Done. Now that `AD_COMPDEP_WATCHED_COL` has its own `AD_MODULE_ID`, the whereclause was simplified from `ad_column_comp_dependency_id in (select ad_column_comp_dependency_id from ad_column_comp_dependency where ad_module_id in (:moduleid))` to a direct `ad_module_id in (:moduleid)`.

src-db/database/sourcedata/AD_REFERENCE.xml
valenvivaldi: No default is needed here. "Normal" is now represented by the *absence* of a value: `Computation_Mode` is nullable and NULL means Normal — all consumers check explicitly for `'S'`/`'V'` and are null-safe, nobody compares against `'N'`. We actually removed the `'N'` list item and the column default entirely, because a mandatory/default value forced the field to be written to the sourcedata of every column in every module (it added ~12k lines to core's `AD_COLUMN.xml`). With NULL=Normal, only stored (`S`) and virtual (`V`) columns carry a value and get exported.

src-db/database/sourcedata/AD_REFERENCE.xml
valenvivaldi: Same reasoning against a column default (it would be written to every column and is meaningless for non-stored ones). But Refresh Mode does need to be present whenever `Computation_Mode='S'`, otherwise `GenerateStoredComputedTriggers` silently skips the column and it never recomputes. Instead of a default we enforce it at save time: added a runtime rule (V18) in the `ColumnStoredComputedHandler` event observer — a stored computed column with no valid Refresh Mode (`S`/`Q`/`M`) is rejected with the new `ETGO_StoredComputedRefreshMode` message. Committed in `79f730f2`, with a unit test covering the null/blank/invalid/valid cases.

Matias-Bernal (approved):

PR #1124, opened 2026-08-11
Issue #1113: Fix empty report filter combos in Firefox 153+

### LINKED ISSUE #1113: ETP-4804: Dimensional Analysis report filters show empty combos in Firefox 153+ (options nested inside subreport placeholder div)
# Error's description
On Firefox 153.0 (and later), filter selectors in Dimensional Analysis reports (JasperReports) — e.g. Organization, Business Partner Group, Product Category, Currency in the "Purchase Order Dimensional Analysis" report — render as empty dropdowns with no selectable options, even though matching data exists.

The same regression was already identified and partially fixed for the "Create Lines From" flow (ESD-1884 / GitHub issue #1102, released in Etendo 25.4.25 / 26.2.6): Firefox 153 stopped treating <option> elements as selectable when they are not direct children of their <select>. The fix introduced reparentComboOptions() in web/js/utils.js, but it was only wired into CreateFrom_Shipment.html and CreateFrom_ShipmentPO.html. It was never propagated to the *FilterJR.html templates under src/org/openbravo/erpCommon/ad_reports/ (40+ files), which render their combos with the exact same problematic pattern:

```
<select name="inpOrg" id="inpOrg" class="ComboKey Combo_TwoCells_width" required="true">
  <option value=""/>
  <div id="reportAD_ORGID"/>
</select>
```
populated server-side via XmlEngine / ComboTableData (ReportPurchaseDimensionalAnalysesJR.java), which leaves the injected <option> elements nested inside the <div id="reportAD_ORGID"/> placeholder instead of as direct children of the <select>.

 # Steps to reproduce the error
1. Use Firefox 153.0 or later.
2. Open the "Purchase Order Dimensional Analysis" report filter (or any other Dimensional Analysis / JasperReport filter under ad_reports).
3. Click any of the filter dropdowns (Organization, Business Partner Group, Product Category, Currency, etc.).
4. Observe that the dropdown shows no selectable options, despite data existing for that field.

https://github.com/user-attachments/assets/7b4f6358-c8fe-4766-9452-ae51f91e5825

 # Expected behavior
Filter dropdowns should display and allow selecting all available options, consistently with Chrome/Chromium-based browsers.

 # Affected Version
Etendo version: 25.4.26 | 26.2.7 (also present since 25.4.25 / 26.2.6, the releases that shipped the partial fix without covering reports)

 # Solution Design (optional)
The server-side liststructure resolution (XmlEngine/XmlDocument, used by ComboTableData.select(...) + xmlDocument.setData(id, "liststructure", ...)) injects the generated <option> elements as children of the placeholder <div id="report...ID"/> instead of replacing that div with sibling <option> nodes inside the <select>. Firefox 153 no longer treats non-direct-child <option> elements as selectable, so the combo appears empty. This is the exact same pattern already fixed for CreateFrom_Shipment.html/CreateFrom_ShipmentPO.html via reparentComboOptions() in web/js/utils.js, but that call was never added to the *FilterJR.html report filter templates.

`..
`
# Other test cases
Given: a user on Firefox 153+ opens "Purchase Order Dimensional Analysis" filter
When: they open the Organization/Currency dropdown
Then: all available options are visible and selectable

Given: a user on Chrome/Edge opens the same filter
When: they open the dropdowns
Then: behavior remains unaffected (no regression)

Given: the fix is applied
When: a different report is opened in Firefox 153+ (e.g. Sales Order Dimensional Analysis, or any other *FilterJR.html)
Then: its filter combos also show options correctly (verifies the fix covers the whole ad_reports family, not just the reported one)

franciscofourcade:
## Summary

- Fixes filter combos rendering as empty dropdowns in Firefox 153+ across the Dimensional
  Analysis / JasperReport filter screens (Organization, Business Partner Group, Product Category,
  Currency, ...), reported for "Purchase Order Dimensional Analysis".
- Root cause: Firefox 153 enabled `dom.lift_select_parser_restrictions.enabled` by default
  ([bug 2019977](https://bugzilla.mozilla.org/show_bug.cgi?id=2019977)). Before that change, the
  HTML parser **dropped** elements that are not allowed inside `<select>`, so the SUBREPORT
  placeholder `<div>` disappeared and the generated `<option>` elements ended up as direct
  children of the `<select>` as a side effect. That silent parser fix-up is what made XmlEngine's
  invalid markup work for the past 20 years. The `<div>` is now kept in the DOM, and Firefox no
  longer includes the nested `<option>` elements in layout: they get a `0x0` box and are never
  painted, so the dropdown renders blank.
- Worth noting for reviewers: `select.options` **still reports the options** and they are
  programmatically selectable — only the rendering is broken. Any check based on
  `select.options.length` will look healthy while the user sees an empty list.
- Fix: adds `reparentAllComboOptions()` to `web/js/utils.js`, which applies the existing
  `reparentComboOptions()` helper (introduced in #1104) to every `<select>` in the document, and
  calls it from the shared `moreOnLoadDoFunctions()` hook. That hook already runs on virtually
  every classic screen via `setWindowTableParentElement()` (`web/js/windowKeyboard.js:170`), so a
  single change covers the whole `ad_reports` family plus `ad_forms`, `ad_process`, `info`,
  the WAD-generated action buttons under `srcAD/` and third-party module templates, with no
  per-screen wiring.
- `ReportRegisterFilter.html` and `ReportPaymentFilter.html` are the only two affected templates
  that do not go through that hook (their `<body onload>` does not call `onLoadDo()`), so they get
  an explicit `reparentAllComboOptions()` call. Both already load `utils.js`.
- The call is wrapped in `try/catch`: `moreOnLoadDoFunctions()` is the first statement of nearly
  every `onLoadDo()`, and an exception there would abort the rest of the onload sequence
  (shortcuts, focus, resize) on every classic screen, including third-party ones.
- The two existing per-combo calls in `CreateFrom_Shipment.html` / `CreateFrom_ShipmentPO.html`
  are left untouched. `reparentComboOptions()` is idempotent, so the double run is a no-op.

## Verification

Measured in Firefox 153.0.3 against Etendo 26.2.7. Because the
dropdown popup is an OS-level widget that cannot be captured, the combos were forced to
`size=6` so the list renders in-page, and each option's `getBoundingClientRect()` was measured.
An option that does not participate in layout (`0x0`) is an option the user cannot see.

Options painted / total:

| Screen | Before | After |
|---|---|---|
| Purchase Order Dimensional Analysis | `inpOrg` 1/9, `inpPartnerGroup2` 1/7, `inpProductCategory2` 1/8, `inpCurrencyId` 1/165 | 9/9, 7/7, 8/8, 165/165 |
| Sales Order Dimensional Analysis | 6 combos, all 1/N | all N/N |
| General Accounting Reports | `inpcAcctSchemaId` 1/3, `inpLevel` 1/5 | 3/3, 5/5 |
| Invoice Vendor Dimensional Analysis | 5 combos, all 1/N | all N/N |
| **Total unpainted options** | **575** | **0** |

Also confirmed: the `selected` attribute survives the re-parenting (`selectedIndex` and `value`
unchanged), running the sweep twice changes nothing (idempotent), and no `SEVERE`/exception
appears in the Tomcat log.

### Verified on the 25.4 line

The same measurement was run against a clean Etendo **25.4.26** environment. The affected
templates there are byte-identical to `origin/release/25.4`, and the served markup is identical to
26.2.7 (same nested `<div id="reportAD_ORGID"><div id="sectionDetail">` inside the `<select>`).

| | Unpainted options across the 4 screens |
|---|---|
| 25.4.26 without the fix | **575** |
| 25.4.26 with this backport | **0** |

## Test plan

- [x] Reproduced the empty dropdowns in Firefox 153.0.3 and confirmed they are populated after the fix
- [x] Verified the fix on three further report families beyond the reported one
- [x] Verified the default selected option is preserved
- [x] Verified idempotency against the combos already wired by #1104

**No automated test added.** This is a browser layout/parsing behaviour and the repository has no
JavaScript test harness (no `package.json`, karma, jest, qunit, Selenium or HtmlUnit); unit tests
here are Java-only. Same rationale as #1104.

## Known technical debt

The underlying cause is server-side: XmlEngine uses the `<SUBREPORT>` anchor as a *container*
instead of replacing it, and `src/org/openbravo/erpCommon/reference/List.srpt` adds a second
wrapper (`<div id="sectionDetail">`). The engine already has a transparent-container mechanism —
`TagValue.print()` suppresses any tag ending in `_TMP`, used for exactly this purpose in
`ModuleBox.srpt` — but applying it would mean touching `List.srpt` plus ~191 anchors across ~92
templates, and would still not cover third-party modules that ship their own templates. Left as
follow-up.

## Jira

ETP-4804

Fixes #1113

### DIFF
--- a/src/org/openbravo/erpReports/ReportPaymentFilter.html
+++ b/src/org/openbravo/erpReports/ReportPaymentFilter.html
@@ -60,7 +60,7 @@ function validate() {
 }
 </script>
 </head>
-    <body leftmargin="0" topmargin="0" onload="reparentAllComboOptions();resizeArea();updateMenuIcon('buttonMenu');setFocusFirstControl();" onresize="resizeArea();">
+    <body leftmargin="0" topmargin="0" marginwidth="0" marginheight="0" onload="resizeArea();updateMenuIcon('buttonMenu');setFocusFirstControl();" onresize="resizeArea();">
 <form method="post" action="ReportPaymentFilter.html" name="frmMain" id="form">
   <input type="hidden" name="Command"></input>
   <input type="hidden" name="inpAuxd" value="" id="functionSumAux1"></input>
@@ -349,5 +349,11 @@ function validate() {
     </tr>
   </table>
 </form>
+<script type="text/javascript">
+  // Firefox 153+ keeps the SUBREPORT placeholder <div> inside the <select>, which
+  // leaves the generated <option> elements out of the layout. Normalize them here:
+  // this screen does not go through onLoadDo() / moreOnLoadDoFunctions().
+  reparentAllComboOptions();
+</script>
 </body>
 </html>
--- a/src/org/openbravo/erpReports/ReportPaymentFilter.html
+++ b/src/org/openbravo/erpReports/ReportPaymentFilter.html
@@ -60,7 +60,7 @@ function validate() {
 }
 </script>
 </head>
-    <body leftmargin="0" topmargin="0" marginwidth="0" marginheight="0" onload="reparentAllComboOptions();resizeArea();updateMenuIcon('buttonMenu');setFocusFirstControl();" onresize="resizeArea();">
+    <body leftmargin="0" topmargin="0" onload="reparentAllComboOptions();resizeArea();updateMenuIcon('buttonMenu');setFocusFirstControl();" onresize="resizeArea();">
 <form method="post" action="ReportPaymentFilter.html" name="frmMain" id="form">
   <input type="hidden" name="Command"></input>
   <input type="hidden" name="inpAuxd" value="" id="functionSumAux1"></input>
--- a/src/org/openbravo/erpReports/ReportPaymentFilter.html
+++ b/src/org/openbravo/erpReports/ReportPaymentFilter.html
@@ -60,7 +60,7 @@ function validate() {
 }
 </script>
 </head>
-    <body leftmargin="0" topmargin="0" marginwidth="0" marginheight="0" onload="resizeArea();updateMenuIcon('buttonMenu');setFocusFirstControl();" onresize="resizeArea();">
+    <body leftmargin="0" topmargin="0" marginwidth="0" marginheight="0" onload="reparentAllComboOptions();resizeArea();updateMenuIcon('buttonMenu');setFocusFirstControl();" onresize="resizeArea();">
 <form method="post" action="ReportPaymentFilter.html" name="frmMain" id="form">
   <input type="hidden" name="Command"></input>
   <input type="hidden" name="inpAuxd" value="" id="functionSumAux1"></input>
--- a/src/org/openbravo/erpReports/ReportRegisterFilter.html
+++ b/src/org/openbravo/erpReports/ReportRegisterFilter.html
@@ -60,7 +60,7 @@ function validate() {
 }
 </script>
 </head>
-    <body leftmargin="0" topmargin="0" onload="reparentAllComboOptions();resizeArea();updateMenuIcon('buttonMenu');setFocusFirstControl();" onresize="resizeArea();">
+    <body leftmargin="0" topmargin="0" marginwidth="0" marginheight="0" onload="resizeArea();updateMenuIcon('buttonMenu');setFocusFirstControl();" onresize="resizeArea();">
 <form method="post" action="ReportRegisterFilter.html" name="frmMain" id="form">
   <input type="hidden" name="Command"></input>
   <input type="hidden" name="inpAuxd" value="" id="functionSumAux1"></input>
@@ -351,5 +351,11 @@ function validate() {
     </tr>
   </table>
 </form>
+<script type="text/javascript">
+  // Firefox 153+ keeps the SUBREPORT placeholder <div> inside the <select>, which
+  // leaves the generated <option> elements out of the layout. Normalize them here:
+  // this screen does not go through onLoadDo() / moreOnLoadDoFunctions().
+  reparentAllComboOptions();
+</script>
 </body>
 </html>
--- a/src/org/openbravo/erpReports/ReportRegisterFilter.html
+++ b/src/org/openbravo/erpReports/ReportRegisterFilter.html
@@ -60,7 +60,7 @@ function validate() {
 }
 </script>
 </head>
-    <body leftmargin="0" topmargin="0" marginwidth="0" marginheight="0" onload="reparentAllComboOptions();resizeArea();updateMenuIcon('buttonMenu');setFocusFirstControl();" onresize="resizeArea();">
+    <body leftmargin="0" topmargin="0" onload="reparentAllComboOptions();resizeArea();updateMenuIcon('buttonMenu');setFocusFirstControl();" onresize="resizeArea();">
 <form method="post" action="ReportRegisterFilter.html" name="frmMain" id="form">
   <input type="hidden" name="Command"></input>
   <input type="hidden" name="inpAuxd" value="" id="functionSumAux1"></input>
--- a/src/org/openbravo/erpReports/ReportRegisterFilter.html
+++ b/src/org/openbravo/erpReports/ReportRegisterFilter.html
@@ -60,7 +60,7 @@ function validate() {
 }
 </script>
 </head>
-    <body leftmargin="0" topmargin="0" marginwidth="0" marginheight="0" onload="resizeArea();updateMenuIcon('buttonMenu');setFocusFirstControl();" onresize="resizeArea();">
+    <body leftmargin="0" topmargin="0" marginwidth="0" marginheight="0" onload="reparentAllComboOptions();resizeArea();updateMenuIcon('buttonMenu');setFocusFirstControl();" onresize="resizeArea();">
 <form method="post" action="ReportRegisterFilter.html" name="frmMain" id="form">
   <input type="hidden" name="Command"></input>
   <input type="hidden" name="inpAuxd" value="" id="functionSumAux1"></input>
--- a/web/js/utils.js
+++ b/web/js/utils.js
@@ -1974,12 +1974,17 @@ function reparentComboOptions(combo) {
 * Combos rendered by XmlEngine keep their <option> elements nested inside the SUBREPORT
 * placeholder <div>, which some browsers (Firefox 153+) no longer expose as selectable
 * options. Running this once the DOM is ready normalizes every combo of the page at once,
-* so no per-screen wiring is required.
+* so no per-screen wiring is required. It never throws, so it is safe to call from any
+* onload handler without guarding the call site.
 */
 function reparentAllComboOptions() {
-  var combos = document.getElementsByTagName("select");
-  for (var i = 0; i < combos.length; i++) {
-    reparentComboOptions(combos[i]);
+  try {
+    var combos = document.getElementsByTagName("select");
+    for (var i = 0; i < combos.length; i++) {
+      reparentComboOptions(combos[i]);
+    }
+  } catch (ignored) {
+    // Combo normalization must never break the onload sequence of any screen
   }
 }
 
@@ -5438,12 +5443,7 @@ function fixIE9WindowBorders() {
 }
 
 function moreOnLoadDoFunctions() {
-  try {
-    reparentAllComboOptions();
-  } catch (e) {
-    // Combo normalization must never break the page onload sequence
-    console.log(e);
-  }
+  reparentAllComboOptions();
   setMDIEnvironment();
   fixIE9WindowBorders();
 }
--- a/web/js/utils.js
+++ b/web/js/utils.js
@@ -1970,6 +1970,20 @@ function reparentComboOptions(combo) {
 }
 
 /**
+* Applies reparentComboOptions to every combo (select) present in the current document.
+* Combos rendered by XmlEngine keep their <option> elements nested inside the SUBREPORT
+* placeholder <div>, which some browsers (Firefox 153+) no longer expose as selectable
+* options. Running this once the DOM is ready normalizes every combo of the page at once,
+* so no per-screen wiring is required.
+*/
+function reparentAllComboOptions() {
+  var combos = document.getElementsByTagName("select");
+  for (var i = 0; i < combos.length; i++) {
+    reparentComboOptions(combos[i]);
+  }
+}
+
+/**
 * Fills a combo with a data from an Array. Allows to set a default selected item, defined as boolean field in the Array.
 * @param {Object} combo A reference to the combo object.
 * @param {Array} dataArray Array containing the data for the combo. The structure of the array must be value, text, selected. Value is value of the item, text the string that will show the combo, an selected a boolean value to set if the item should appear selected.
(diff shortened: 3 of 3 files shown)


fran-roig (approved):

src/org/openbravo/erpReports/ReportPaymentFilter.html
```diff
@@ -60,7 +60,7 @@
 }
 </script>
 </head>
-    <body leftmargin="0" topmargin="0" marginwidth="0" marginheight="0" onload="resizeArea();updateMenuIcon('buttonMenu');setFocusFirstControl();" onresize="resizeArea();">
+    <body leftmargin="0" topmargin="0" onload="reparentAllComboOptions();resizeArea();updateMenuIcon('buttonMenu');setFocusFirstControl();" onresize="resizeArea();">
```
valenvivaldi: **Unrelated change — please restore.** This line also drops `marginwidth="0" marginheight="0"`, which has nothing to do with the combo fix.

These templates are XHTML Transitional documents rendered inside frames, and those legacy attributes are what zero out the body margin on several engines. The removal is also inconsistent with the rest of the tree: **300 files still carry them**, including the neighbouring `RptC_Bpartner.html` in this same directory.

The diff here should be just the added `reparentAllComboOptions();` at the head of `onload`.

Separately, a consistency note: the `try/catch` in `moreOnLoadDoFunctions()` is justified by "an exception would abort the rest of the onload sequence" — but this call site is unguarded, so if it throws here, `resizeArea()` and `setFocusFirstControl()` are lost on this screen. The same rationale argues for the same protection in both places.

src/org/openbravo/erpReports/ReportRegisterFilter.html
```diff
@@ -60,7 +60,7 @@
 }
 </script>
 </head>
-    <body leftmargin="0" topmargin="0" marginwidth="0" marginheight="0" onload="resizeArea();updateMenuIcon('buttonMenu');setFocusFirstControl();" onresize="resizeArea();">
+    <body leftmargin="0" topmargin="0" onload="reparentAllComboOptions();resizeArea();updateMenuIcon('buttonMenu');setFocusFirstControl();" onresize="resizeArea();">
```
valenvivaldi: Same as in `ReportPaymentFilter.html`: `marginwidth="0" marginheight="0"` is removed here too, unrelated to the fix — please restore. And the unguarded call has the same `try/catch` asymmetry noted there.

(Confirmed `utils.js` is loaded in both templates at line 34, so the call resolves.)

valenvivaldi (commented):
Reviewed the change and cross-checked the claims in the description against the actual template/JS sources. The approach is well scoped: hooking the sweep into `moreOnLoadDoFunctions()` instead of touching ~92 templates or `List.srpt` is the right trade-off, and the reasoning about the Firefox 153 parser change matches what the markup does.

**What I independently verified (all confirmed):**

1. **The coverage claim is exact.** I scanned every `*.html` under `src/`, `srcAD/`, `modules_core/` and `modules/` for templates containing a `<select>` with a wrapper (`<div>`/`<SUBREPORT>`) whose `<body onload>` does *not* route through `onLoadDo()`. The result is exactly `ReportPaymentFilter.html` and `ReportRegisterFilter.html` — the two templates patched here. There are also zero templates that carry a combo and have no `onload` hook at all. So the two explicit calls are the complete remainder, not an estimate.

2. **Deleting the wrapper `<div>` is safe.** This was my main worry, since `reparentComboOptions()` removes *every* non-`OPTION`/`OPTGROUP` element child and those wrappers carry ids (`reportAD_ORGID`, `sectionDetail`, `sectionModuleComboDetail`, …). I extracted all **135 wrapper ids** that appear inside a `<select>` across the repo and grepped each one against `web/`, `src/`, `srcAD/` and `modules_core/`: **none is referenced from JS**. Nothing looks the placeholder up after onload, so the removal cannot break a post-load code path.

3. **Post-load combo refresh is unaffected.** The AJAX path goes through `fillCombo()` (`utils.js:2840`), which builds options with `new Option()` and assigns via `combo.options[i]` — they land as direct children, so a refreshed combo does not reintroduce the nesting.

4. **Hook placement is correct.** The sweep is the first statement of `moreOnLoadDoFunctions()`, which is itself the first statement of `onLoadDo()`, so combos are normalized before `setFocusFirstControl()`/resize run.

**One coordination item:** the description documents the 575 → 0 measurement on a clean 25.4.26 environment, but I don't see a companion PR against `release/25.4`. The convention in this repo is a paired PR (cf. #1116 → `main` / #1115 → `release/25.4` for the same ETP-4765). If the 25.4 line should get this, that second PR is missing; if it deliberately should not, worth stating so in the body so it doesn't read as an oversight.

Inline comments below: one small revert request in the two templates, plus two non-blocking observations. Nothing here challenges the fix itself.

web/js/utils.js
```diff
@@ -1969,6 +1969,20 @@ function reparentComboOptions(combo) {
...
 }
 
+/**
+* Applies reparentComboOptions to every combo (select) present in the current document.
+* Combos rendered by XmlEngine keep their <option> elements nested inside the SUBREPORT
+* placeholder <div>, which some browsers (Firefox 153+) no longer expose as selectable
+* options. Running this once the DOM is ready normalizes every combo of the page at once,
+* so no per-screen wiring is required.
+*/
+function reparentAllComboOptions() {
```
valenvivaldi, line 1980: **Non-blocking, but worth a spot check: option *order* changes in a handful of combos.**

Where the wrapper `<div>` precedes a direct-child `<option>` in the template, the resulting order differs from what Firefox <153 used to show. The old parser hoisted the generated options *before* the static one; re-parenting keeps the static option where it is and appends the nested ones after it, so the static option moves to the front.

Templates with that shape:

- `src/org/openbravo/erpCommon/ad_forms/RequisitionToOrderCreate.html` — `inpPriceListId`, `inpWarehouse`
- `src/org/openbravo/erpCommon/ad_forms/ModuleManagementSettings.html` — `inpModule`
- `src/org/openbravo/erpCommon/ad_forms/ShowSessionPreferences.html` — `inpTheme`
- `src/org/openbravo/erpCommon/info/Location_F1.html` — `inpCRegionId`

In practice this is cosmetic and arguably more correct (the blank option ends up first, which is the usual convention), and `selectedIndex`/`value` are preserved. But that is exactly why the verification described in the PR would not surface it: checking the selected value does not check ordering. A quick look at those four screens would close the gap.

web/js/utils.js
```diff
@@ -5424,6 +5438,12 @@ function fixIE9WindowBorders() {
 }
 
 function moreOnLoadDoFunctions() {
+  try {
+    reparentAllComboOptions();
+  } catch (e) {
+    // Combo normalization must never break the page onload sequence
+    console.log(e);
```
valenvivaldi: Nit: `console.log(e)` inside the catch is a residual risk in a browser without `console` — it would throw from within the handler that exists precisely to guarantee onload never breaks. It is consistent with existing usage in this file (lines 511, 543, 588), so not worth changing on its own; flagging only because this particular `catch` is the one place where the guarantee is the point.

src/org/openbravo/erpReports/ReportPaymentFilter.html
franciscofourcade: Both points are fair, and the second one is a real defect in the diff — thanks.

**Unrelated change:** agreed, and I've taken the removal out. Rather than restoring the attributes and re-triggering the quality gate (the 4 `Web:S1827` issues Sonar reported were exactly these two attributes on this line — pre-existing, but counted as new code because the line was modified), the call no longer touches this line at all. Line 63 is back to its original form, `marginwidth="0" marginheight="0"` included, and the call moved to a small script block before `</body>`, which runs once the server-rendered `<select>` elements are in the DOM. The attributes are indeed pervasive — 197 files in the source tree carry them — so removing them here was inconsistent regardless of Sonar.

**try/catch asymmetry:** you're right, and the inconsistency was mine. I justified the guard by "an exception would abort the rest of the onload sequence" and then added an unguarded call to an `onload` where precisely that would happen. Fixed by moving the guard inside `reparentAllComboOptions()` instead of wrapping one call site: both entry points are now protected, and `moreOnLoadDoFunctions()` goes back to a plain call. That also removes the `console.log` you flagged in the other thread.

The net diff against `main` is now +32/-0 across the three files — nothing pre-existing is modified.

src/org/openbravo/erpReports/ReportRegisterFilter.html
franciscofourcade: Same resolution as in `ReportPaymentFilter.html`: line 63 is restored to its original form, `marginwidth="0" marginheight="0"` included, and the call moved to a script block before `</body>` so this line is no longer part of the diff. The `try/catch` asymmetry is fixed by moving the guard inside `reparentAllComboOptions()`, which covers both call sites.

Thanks for confirming the `utils.js` load at line 34.

web/js/utils.js
franciscofourcade: Good catch to check, but I could not reproduce an order change — and the templates you list do have that tag sequence, so it's worth showing why.

In all of those selects the `<option>` elements that appear after the `<div>` in the template are either **inside** the wrapper (they are the SECTION detail template) or design placeholders marked `id="discard"` / `class="discard"`, which XmlEngine strips. So in the rendered output there is no direct-child `<option>` after the wrapper. For example `info/DebtPayment.html` renders as `option, div, div, option, option, ...` — a blank option, then the wrapper containing the generated ones. Re-parenting appends the generated options after that blank option, which is the same position the old parser hoisted them to.

Rather than eyeballing it, I compared against the pre-153 parser directly by launching Firefox 153 with `dom.lift_select_parser_restrictions.enabled=false` (which restores the old behaviour and makes the sweep a no-op, since the options are already direct children) and diffing the option text sequence, count and `selectedIndex` against 153 + the fix:

| Screen | Combo | legacy vs 153+fix |
|---|---|---|
| `ad_forms/ShowSessionPreferences.html` | `reportTheme_S` | n=2/2, selectedIndex=0/0, identical order |
| `info/DebtPayment.html` | `inpCPaymentRuleId` | n=12/12, selectedIndex=0/0, identical order |

Zero differences. I measured these two because they are the ones reachable directly by URL; for the rest I verified the markup shape, which is the same in all 8 cases.

One addition to your list, in case it's useful: the same template shape also appears in `info/Location_F1.html` (`inpCCountryId`, besides the `inpCRegionId` you found), `info/DebtPayment.html` (`inpCPaymentRuleId`) and `ad_forms/ShowSessionVariables.html` (`inpWindow`).

Leaving this thread open for you to close, since it's your call whether the measurement covers the concern.

web/js/utils.js
franciscofourcade: Agreed, and the guarantee being the point here is exactly why it's worth changing. Resolved by moving the `try/catch` inside `reparentAllComboOptions()` and dropping the `console.log` altogether, using the `catch (ignored) {}` idiom already used in this file (lines 325, 642, 723, 777). That removes the `console` dependency from the one handler whose job is to never throw, and it also protects the second call site in the report templates, which was unguarded.

PR #1127, opened 2026-08-12
Issue #1121: Fix blank Email options popup when adding an attachment

### LINKED ISSUE #1121: ETP-4868: When "Add Attachment" is selected, the Email options popup goes blank (empty 200 OK response)
## Error's description

When a user opens the "Email options" popup from a Sales Invoice (or any document sent by email) and clicks "Add Attachment" after selecting a file, the popup goes completely blank. Network inspection shows a `POST /invoices/PrintOptions.html` request with `Command=ADD` (multipart/form-data body containing the selected file) that returns `200 OK` with `Content-Length: 0` and an empty body. No exception is logged at any layer (Apache/mod_jk, Tomcat, or the Etendo application logger), because the failure is not an exception — it's a silent no-op in the command dispatch logic.

Image

## Steps to reproduce the error

1. Open a completed Sales Invoice.
2. Click the **Email** icon — the "Email options" popup loads correctly with all fields (from/to/subject/template/attachments).
3. Click **Select file** and choose any file.
4. Click **Add Attachment**.
5. The popup goes blank. DevTools Network tab shows the POST to `/invoices/PrintOptions.html` (`Command=ADD`) returning `200 OK`, `Content-Length: 0`.

https://github.com/user-attachments/assets/011aff21-5b2d-4284-bc0d-750aad5e4411

## Expected behavior

The selected file should be added to the "Attached Documents" list and the popup should re-render with the same form fields plus the newly added attachment.

## Affected Version

- **Etendo version:** introduced in `26.1.10`, confirmed present through `26.2.7` (verified directly against GitHub tags: `PrintControllerCommandHandler.java` does **not** exist in tag `26.1.9` (404), but **does** exist in tag `26.1.10` (200) with this exact bug already present, byte-for-byte identical to `26.2.7`).
- **NOT present in 25.x:** the file does not exist in `25.4.26` (latest 25.x) — the 25.x line never received the refactor that introduced this bug.
- **Module/Bundle affected:** Etendo Core (not a module) — `org.openbravo.erpCommon.utility.reporting.printing`
- **Related to** #1049 ("Email sending fails with TypeError when selecting invoices from multiple customers", released in `26.1.10`): this bug is a side-effect regression introduced by that fix, not the same defect. That fix refactored `PrintController.post()`'s monolithic ADD-command handling (a simple `if (print.html) {...} else {...}` which always resolved to one branch) into the current `PrintControllerCommandHandler.openOptionsPage()` with two independent `if`s and no fallback — leaving the gap for the `PrintOptions.html` path described below.

### Root cause

Confirmed in `src/org/openbravo/erpCommon/utility/reporting/printing/PrintControllerCommandHandler.java:361-372`, method `openOptionsPage()`:

```java
private void openOptionsPage() throws IOException, ServletException, ReportingException {
  String docIdsForPage = PrintControllerDocumentHelper.getCommaSeparatedString(context.documentIds);
  if (isPrintPath()) {
    controller.createPrintOptionsPage(request, response, vars, context.documentType,
        docIdsForPage, context.reports);
    return;
  }
  if (isSendPath()) {
    controller.createEmailOptionsPage(request, response, vars,
        docIdsForPage, context.reports, context.checks);
  }
}

with the path helpers (lines 392-402):

private boolean isPrintPath()        { return request.getServlf("print.html") != -1; }
private boolean isPrintOptionsPath() { return request.getServlf("printoptions.html") != -1; }
private boolean isSendPath()         { return request.getServlf("send.html") != -1; }

The "Add Attachment" button submits Command=ADD to /invoices/Pth:

- isPrintPath() → false (the literal substring "print.html" is.html" — the characters after "print" are "options", not".html")
- isSendPath() → false (no "send.html" substring)

Neither if matches, so openOptionsPage() returns without ever e or createEmailOptionsPage — nothing is ever written to theresponse, producing the empty 200 OK. Note that isPrintOptionsrectly detects this exact path, but it is only used invalidateSenderConfiguration() (line 376) — never wired into op

Proposed fix

if (isSendPath() || isPrintOptionsPath()) {
  controller.createEmailOptionsPage(request, response, vars, c
      docIdsForPage, context.reports, context.checks);
}

(or restore an if/else structure so every path resolves to one Investigation notes: bundle/module set, SMTP configuration, aned out as causes during live investigation on demo.etendo.cloud (26.2.7, Tomcat 9.0.98 + Apache/mod_jk). The bug depends pureld reproduces consistently once a real file is attached via the popup, regardless of environment, as long as the request path is exactly PrintOptions.html.

Secondary/cosmetic finding (not the root cause but worth fixing together): EmailOptions.html has malformed/unbalanced HTML tags (div/tr/td) in the
"Attached Documents" section, around lines 960-1035.

Other test cases

- Given: A completed Sales Order (or Shipment/Payment) with a
When: The user opens its Email options popup and clicks "Add Aed
Then: The same blank-popup / empty response bug should reproduce, since PrintControllerCommandHandler is shared across all document types, not just
invoices.
- Given: The Print options popup (not Email) for any document
When: The user triggers an "ADD"-type action within that popup
Then: This should not be affected, since the request path thertly matches isPrintPath().

Related issues

- #1049 — root cause of this regression (26.1.10)
- ETP-4868 / ESD-1902 — internal tracking

### PR DESCRIPTION
**Pull request by franciscofourcade**
Fixes #1121 — ETP-4868

## Root cause

`PrintInvoices` (and every other print controller) is mapped to three URLs: `/invoices/send.html`, `/invoices/print.html` and `/invoices/PrintOptions.html`. The Email options popup is opened through `send.html`, but its own form posts back to `PrintOptions.html`:

```html
<form id="form" method="POST" action="PrintOptions.html" name="frmMain" enctype="multipart/form-data">
```

So the `ADD` command issued by "Add Attachment" arrives on `PrintOptions.html`, where `openOptionsPage()` matched neither branch:

- `isPrintPath()` → `false`, because the literal substring `print.html` is not contained in `printoptions.html` (the characters after `print` are `options`, not `.html`)
- `isSendPath()` → `false`, no `send.html` substring

The method returned without writing anything to the response, producing the empty `200 OK` and the blank popup. No exception anywhere, because it was a silent no-op rather than a failure.

This is a regression from the ETP-4197 refactor (#1049, released in 26.1.10). Before it, the `ADD` command in `PrintController.post()` was an `if/else` with a fallback:

```java
} else if (vars.commandIn("ADD")) {
  if (request.getServletPath().toLowerCase().indexOf(PRINT_PATH) != -1) {
    createPrintOptionsPage(...);
  } else {
    createEmailOptionsPage(...);
  }
```

The refactor turned that into two independent `if`s with no fallback.

## Fix

Restore the `if/else` shape in `PrintControllerCommandHandler.openOptionsPage()`: `print.html` renders the print options page, **every other path** falls back to the email options page. This was preferred over adding `isPrintOptionsPath()` as a third condition because it removes the whole class of failure — an unmatched path can no longer leave the response empty — and it matches the pre-regression behaviour exactly.

`isSendPath()` becomes unused and is removed (`isPrintOptionsPath()` stays, it is still used by `validateSenderConfiguration()`).

## Tests

`PrintControllerCommandHandlerTest` had **encoded the bug**: `testHandle_addCommand_otherPath_callsNeither()` asserted that neither page builder was called. That expectation was replaced with correct coverage:

- `testHandle_addCommand_printOptionsPath_callsCreateEmailOptionsPage` — the exact regression
- `testHandle_addCommand_printOptionsPath_salesOrder_callsCreateEmailOptionsPage` — the handler is shared across document types
- `testHandle_addCommand_otherPath_fallsBackToEmailOptionsPage` — the response is never left empty
- `testIsPrintPath_printOptionsHtml_returnsFalse` — documents the substring trap that caused this

The three new `ADD` tests **fail against the pre-fix code** and pass with it. The whole `reporting.printing` test package (13 classes, 181 tests) passes.

## Manual verification

Verified on a local 26.2.7 instance whose core sources were confirmed byte-for-byte identical to `main` before patching:

1. Reproduced the blank popup on the unpatched instance.
2. Deployed the fixed class, restarted, repeated the flow with a real file attached → the popup re-renders with all fields plus the new entry in "Attached Documents".
3. Checked the Print options popup (`print.html` path) still renders correctly — no regression on that branch.

## Also included

Two malformed-markup fixes in the "Attached Documents" block of `EmailOptions.html`: an unclosed `<th>` and a `value="""` attribute. The deprecated `width` attribute on the same line was converted to an inline style so the change does not introduce a new Web-analyzer violation on a modified line.

**Deliberately not touched:** the unbalanced `div`/`tr`/`td` nesting around the same block. Those `<div id="sectionDetail">` / `<div id="sectionDetail2">` elements are declared as `<SECTION>` in `EmailOptions.xml`, so they are XmlEngine section markers and their boundaries define which markup repeats per row. Rebalancing them would change repetition semantics, not just formatting — it belongs in its own change with its own validation.

### DIFF
--- a/src-test/src/org/openbravo/erpCommon/utility/reporting/printing/PrintControllerCommandHandlerTest.java
+++ b/src-test/src/org/openbravo/erpCommon/utility/reporting/printing/PrintControllerCommandHandlerTest.java
@@ -61,6 +61,8 @@ public class PrintControllerCommandHandlerTest {
   private static final String DOC_ID = "DOC001";
   private static final String SUFFIX_DOCUMENTS = ".Documents";
   private static final String PATH_SEND_HTML = "/send.html";
+  private static final String PATH_PRINT_OPTIONS_HTML = "/invoices/PrintOptions.html";
+  private static final String METHOD_IS_PRINT_PATH = "isPrintPath";
 
   /** Context constructor stores all provided fields. */
   @Test
@@ -309,11 +311,58 @@ public class PrintControllerCommandHandlerTest {
   }
 
   /**
-   * ADD command on a path that is neither print nor send calls neither page builder.
+   * ADD command on the PrintOptions.html path delegates to createEmailOptionsPage. This is the
+   * path the "Add Attachment" button of the email options popup posts to, and it must render the
+   * email options page again instead of leaving the response empty.
    * @throws Exception if handle() propagates an unexpected error
    */
   @Test
-  public void testHandle_addCommand_otherPath_callsNeither() throws Exception {
+  public void testHandle_addCommand_printOptionsPath_callsCreateEmailOptionsPage()
+      throws Exception {
+    PrintController ctrl = mock(PrintController.class);
+    HttpServletRequest req = mock(HttpServletRequest.class);
+    HttpServletResponse resp = mock(HttpServletResponse.class);
+    VariablesSecureApp vars = mock(VariablesSecureApp.class);
+    when(vars.commandIn("ADD")).thenReturn(true);
+    when(req.getServletPath()).thenReturn(PATH_PRINT_OPTIONS_HTML);
+
+    PrintControllerCommandHandler handler = new PrintControllerCommandHandler(
+        ctrl, req, resp, vars, makeContext(new String[]{ DOC_ID }, DocumentType.SALESINVOICE));
+    handler.handle();
+
+    verify(ctrl).createEmailOptionsPage(any(), any(), any(), any(), any(), any(), any());
+    verify(ctrl, never()).createPrintOptionsPage(any(), any(), any(), any(), any(), any());
+  }
+
+  /**
+   * ADD command on a document type other than invoice also reaches the email options page, since
+   * the command handler is shared by every printable document.
+   * @throws Exception if handle() propagates an unexpected error
+   */
+  @Test
+  public void testHandle_addCommand_printOptionsPath_salesOrder_callsCreateEmailOptionsPage()
+      throws Exception {
+    PrintController ctrl = mock(PrintController.class);
+    HttpServletRequest req = mock(HttpServletRequest.class);
+    HttpServletResponse resp = mock(HttpServletResponse.class);
+    VariablesSecureApp vars = mock(VariablesSecureApp.class);
+    when(vars.commandIn("ADD")).thenReturn(true);
+    when(req.getServletPath()).thenReturn("/orders/PrintOptions.html");
+
+    PrintControllerCommandHandler handler = new PrintControllerCommandHandler(
+        ctrl, req, resp, vars, makeContext(new String[]{ DOC_ID }, DocumentType.SALESORDER));
+    handler.handle();
+
+    verify(ctrl).createEmailOptionsPage(any(), any(), any(), any(), any(), any(), any());
+  }
+
+  /**
+   * ADD command on any path that is not print.html falls back to the email options page, so the
+   * response is never left empty.
+   * @throws Exception if handle() propagates an unexpected error
+   */
+  @Test
+  public void testHandle_addCommand_otherPath_fallsBackToEmailOptionsPage() throws Exception {
     PrintController ctrl = mock(PrintController.class);
     HttpServletRequest req = mock(HttpServletRequest.class);
     HttpServletResponse resp = mock(HttpServletResponse.class);
@@ -326,7 +375,7 @@ public class PrintControllerCommandHandlerTest {
     handler.handle();
 
     verify(ctrl, never()).createPrintOptionsPage(any(), any(), any(), any(), any(), any());
-    verify(ctrl, never()).createEmailOptionsPage(any(), any(), any(), any(), any(), any(), any());
+    verify(ctrl).createEmailOptionsPage(any(), any(), any(), any(), any(), any(), any());
   }
 
   /**
@@ -456,7 +505,7 @@ public class PrintControllerCommandHandlerTest {
     HttpServletRequest req = mock(HttpServletRequest.class);
     when(req.getServletPath()).thenReturn("/print.html");
 
-    assertTrue(invokeBooleanReflection(req, "isPrintPath"));
+    assertTrue(invokeBooleanReflection(req, METHOD_IS_PRINT_PATH));
   }
 
   /**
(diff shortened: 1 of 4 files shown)


etendobot (approved):

etendobot (approved):

franciscofourcade:
### SnykCode alert 3828 (`java/RegexInjection`) dismissed as a false positive

The 280-char limit on the dismissal comment does not fit the reasoning, so it is recorded here.

**Where the alert points.** `PrintControllerCommandHandler.java:372` — the `controller.createEmailOptionsPage(...)` call. That line was not added by this PR; it was dedented out of the `if (isSendPath())` block it used to sit in, which is enough for Snyk to treat it as changed and re-report a flow that already existed.

**Why it does not hold.** The sink is `VariablesBase.transformNumber()`:

```java
// src-core/src/org/openbravo/base/VariablesBase.java:813-827
String groupSeparator   = getSessionValue("#GROUPSEPARATOR|"   + DEFAULT_FORMAT_NAME);
String decimalSeparator = getSessionValue("#DECIMALSEPARATOR|" + DEFAULT_FORMAT_NAME);
...
value = value.replaceAll(groupSeparator, "");
```

In `replaceAll(regex, replacement)` the **pattern** is `groupSeparator`. It is written once per session at login, from the server-side `Format.xml`:

```java
// src/org/openbravo/base/secureApp/LoginUtils.java:530-533
vars.setSessionValue("#DecimalSeparator|" + strNumberName,
    NumberElement.getAttributes().getNamedItem("decimal").getNodeValue());
vars.setSessionValue("#GroupSeparator|" + strNumberName,
    NumberElement.getAttributes().getNamedItem("grouping").getNodeValue());
```

That is the only write to those session keys in the codebase — no request parameter reaches them. The HTTP-controlled data flows into `value`, the subject string being scanned, not into the pattern. A ReDoS requires control over the pattern, so the precondition for this rule is not met.

**Alert 3829 (`Trust Boundary Violation`, note) is left open.** It is not a false positive and it does not fail the check. Its sink is `VariablesBase.setSessionObject` (`src-core`), untouched by this PR.

Separately, `PrintController.createEmailOptionsPage()` now builds the `pocData` session key with `sanitizeDocumentIdentifier()` instead of `normalizeDocumentId()`, matching the sanitizer `PrintControllerCommandHandler` already uses to read that same key back. That was a real write/read key mismatch found while tracing this alert, not a Snyk workaround.

etendobot (approved):

fran-roig (approved):

